PR #11, opened 2015-04-13
Jko/futurize

jkodumal:
This should make our Python SDK compatible with Python 2 and 3. It's currently tested against python 2.6.8 and 3.3.3 (chosen arbitrarily).


pkaeding:
lgtm

PR #17, opened 2015-09-08
Support for server-sent events

jkodumal:


ldclient/__init__.py
```diff
@@ -43,16 +47,121 @@ def emit(self, record):
...
+            elif f is None:
+                f = {'deleted': True, 'version': version}
+                self._features[key] = f
+        finally:
+            self._lock.unlock()        
+
+    def upsert(self, key, feature):
+        try:
+            self._lock.lock()
+            f = self._features[key]
```
pkaeding: I think this will throw a `KeyError` if the feature is not there (instead of `None`, which we are checking for on the next line).

    f = self._features.get(key, None)

I think should do what we want.

jkodumal: Good catch, should be fixed in all 3 places (get, upsert, delete)

ldclient/rwlock.py
```diff
@@ -0,0 +1,39 @@
...
+        try:
+            self._readers += 1
+        finally:
+            self._read_ready.release(  )
+
+    def runlock(self):
+        """ Release a read lock. """
+        self._read_ready.acquire()
+        try:
+            self._readers -= 1
```
pkaeding: we should ensure that the number of readers is never negative.

ldclient/rwlock.py
```diff
@@ -0,0 +1,39 @@
+import threading
+
+class ReadWriteLock:
```
pkaeding: note: need to attribute the source of this

pkaeding:
lgtm, besides the minor issues pointed out above.

mrdon:
Is this going in today or should I do the twisted port off this branch?

jkodumal:
I think it's ready to go in today. I can make the improvements to the poached RWLock impl as a bug fix. This is passing integration tests, so it's good to go.

ldclient/rwlock.py
jkodumal: Filed as: https://github.com/launchdarkly/python-client/issues/20

Issue #18, opened 2015-09-08
Twisted support


mrdon:
Would be nice to have twisted support

jkodumal:
Twisted python is now supported in version 0.18.0 as an optional dependency. Thanks @mrdon for the contributions!

PR #19, opened 2015-09-08
Add twisted support #18

mrdon:


mrdon:
The test failures seem to be because twisted dropped 2.6 support... Wouldn't suppose you could too?

jkodumal:
I think we could-- none of our Python customers are on 2.6.

ldclient/__init__.py
```diff
@@ -161,6 +161,7 @@ def _send(self, event):
         if self._queue.full():
             log.warning("Event queue is full-- dropped an event")
         else:
+            print("putting in {}".format(event))
```
jkodumal: Scrap the debug code?

ldclient/twisted.py
```diff
@@ -0,0 +1,179 @@
...
+        self._looping_call = task.LoopingCall(self._consume)
+        self._looping_call.start(5)
+
+    def stop(self):
+        self._looping_call.stop()
+
+    def is_alive(self):
+        return self._looping_call is not None and self._looping_call.running
+
+    def flush(self):
```
jkodumal: If flush is called on an empty queue, it will block until the next time `_consume` has actual work to do. This could be longer than 5 seconds if the queue is empty. 

I also think `flush` should force an immediate request with the current queue contents, rather than merely acting as a synchronization point indicating that the next `_consume` cycle has completed.

ldclient/twisted_sse.py
```diff
@@ -0,0 +1,170 @@
...
+        return ClientContextFactory.getContext(self)
+
+
+class TwistedSSEClient(object):
+    def __init__(self, url, headers, verify, on_event):
+        self.url = url
+        self.verify = verify
+        self.headers = headers
+        self.on_event = on_event
+        self.on_error_retry = 30
```
jkodumal: Feels a bit long. I would think something like 2 seconds is more reasonable on a non-200 response to handle network blips.

ldclient/__init__.py
```diff
@@ -340,7 +358,10 @@ def _headers(api_key):
     return {'Authorization': 'api_key ' + api_key, 'User-Agent': 'PythonClient/' + __version__, 'Content-Type': "application/json"}
 
 def _stream_headers(api_key):
-    return {'Authorization': 'api_key ' + api_key, 'User-Agent': 'PythonClient/' + __version__, 'Accept': "text/event-stream"}
+    return {'Authorization': 'api_key ' + api_key,
+            'User-Agent': 'PythonClient/' + __version__,
```
jkodumal: It would be nice to set the `User-Agent` to `TwistedPythonClient` since the codebases are so different

ldclient/__init__.py
```diff
@@ -135,27 +139,19 @@ def __init__(self, api_key, config):
...
         self._config = config
-        self._store = InMemoryFeatureStore()
+        self._store = config._feature_store_class()
+        self._running = False
 
     def run(self):
         log.debug("Starting stream processor")
+        self._running = True
         hdrs = _stream_headers(self._api_key)
         uri = self._config._stream_uri + "/"
```
jkodumal: We're making one minor change to the streaming SDKs-- rather than subscribe to the root resource `https://stream.launchdarkly.com/`, we're going to subscribe to `https:/stream.launchdarkly.com/features`.

We're doing this because we're going to start adding other things to `stream.ld.com` that are not relevant to the SDKs, and the root resource is intended to subscribe you to everything. 

Besides changing the URI here, you'll also need to change the event names (e.g. `put/features` becomes `put`). Here's the change we made to ruby, as a reference:

https://github.com/launchdarkly/ruby-client/commit/e2aeae5ac21d6bf9a468b271515d3497dda2030d

I was going to do this myself, but since your PR is still out I'd either have to fork your fork, or make the change on master and make you deal with the conflict. I figure just having you do it is easiest.

jkodumal:
Passes all integration tests with flying colors.

PR #23, opened 2015-09-16
PEP-8 + big refactor

mrdon:
Ok, so I got a little carried away here, but I think it'll be much more maintainable and easy to add in new extensions.  I wanted to isolate logic in LDClient and avoid extending classes to add my twisted bits and probably soon asyncio support.  I had to split a lot up into smaller modules to, among other things, avoid circular dependencies.  The public API should be the same, however.


ldclient/client.py
```diff
@@ -0,0 +1,228 @@
...
+    def flush(self):
+        if self._offline:
+            return
+        self._check_consumer()
+        return self._consumer.flush()
+
+    def get_flag(self, key, user, default=False):
+        return self.toggle(key, user, default)
+
+    def toggle(self, key, user, default=False):
```
mrdon: This function is the one significant change here.  The implementation is simpler in that it isolates the actual retrieval from the rest, but more complicated in the callback function.  That func is necessary to support returning defers/tasks for async and the actual value for normal usage.

ldclient/client.py
```diff
@@ -0,0 +1,228 @@
...
+        check_uwsgi()
+        self._api_key = api_key
+        self._config = config or Config.default()
+        self._session = CacheControl(requests.Session())
+        self._queue = queue.Queue(self._config.capacity)
+        self._consumer = None
+        self._offline = False
+        self._lock = Lock()
+
+        self._store = config.feature_store_class()
```
mrdon: Notice the client manages the store now

ldclient/interfaces.py
```diff
@@ -0,0 +1,136 @@
...
+    # noinspection PyMethodMayBeStatic
+    def is_alive(self):
+        """
+        Returns whether the operation is alive or not
+        :rtype: bool
+        """
+        return True
+
+
+class StreamProcessor(BackgroundOperation):
```
mrdon: No methods here as we really don't need any.  Perhaps we should pass in the store more explicitly outside the constructor?

ldclient/client.py
```diff
@@ -0,0 +1,228 @@
...
+    def toggle(self, key, user, default=False):
+
+        default = self._config.get_default(key, default)
+
+        if self._offline:
+            return default
+
+        def cb(feature):
+            if feature is None:
+                return default
```
jkodumal: This should send a flag request event too.

ldclient/client.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if self._config.stream and self._store.initialized:
+            return cb(self._store.get(key))
+        else:
+            # noinspection PyBroadException
+            try:
+                return self._feature_requester.get(key, cb)
+            except Exception:
+                log.exception('Unhandled exception. Returning default value for flag.')
+                return default
```
jkodumal: same here

ldclient/interfaces.py
```diff
@@ -0,0 +1,136 @@
...
+    @abstractmethod
+    def flush(self):
+        """
+        Flushes any outstanding events immediately.
+        """
+
+
+class FeatureRequester(object):
+    """
+    Consumes events from the client and sends them to LaunchDarkly
```
jkodumal: Copypasta

ldclient/interfaces.py
```diff
@@ -0,0 +1,136 @@
...
+    #     """
+    #     Gets feature data
+    #
+    #     :param key: The feature flag key
+    #     :type key: str
+    #     :rtype: dict
+    #     """
+
+
+class Consumer(BackgroundOperation):
```
jkodumal: Maybe we should call this `EventConsumer` now that we have a proliferation of interfaces.

ldclient/interfaces.py
```diff
@@ -0,0 +1,136 @@
...
+class FeatureRequester(object):
+    """
+    Consumes events from the client and sends them to LaunchDarkly
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def get(self, key, callback):
+        """
+        Flushes any outstanding events immediately.
```
jkodumal: Copypasta

ldclient/client.py
```diff
@@ -0,0 +1,229 @@
...
+    @property
+    def initialized(self):
+        try:
+            self._lock.rlock()
+            return self._initialized
+        finally:
+            self._lock.runlock()
+
+
+class LDClient(object):
```
jkodumal: I preferred having `TwistedLDClient`. Most of our customers don't need to set additional config options, so being able to instantiate a client in one line via:

```
client = TwistedLDClient("API_KEY")
```

is nice.

testing/test_integration_twisted.py
```diff
@@ -1,8 +1,10 @@
 import logging
-from ldclient.twisted import TwistedLDClient, TwistedConfig
+from ldclient.client import LDClient
+from ldclient.twisted_impls import TwistedConfig
```
jkodumal: I also liked the aesthetics of importing `ldclient.twisted` instead of `ldclient.twisted_impls`

demo/demo_twisted.py
```diff
@@ -1,11 +1,12 @@
...
 from twisted.internet import task, defer
 
+
 @defer.inlineCallbacks
-def main(reactor):
-    apiKey = 'whatever'
-    client = TwistedLDClient(apiKey)
+def main(_):
+    api_key = 'whatever'
+    client = TwistedLDClient(api_key)
```
jkodumal: Given the current design, this won't work-- this needs to be an `LDClient` with a `TwistedConfig`

jkodumal:
lgtm once those minor comments are addressed.

mrdon:
Comments addressed and now with end-to-end LDD tests

PR #26, opened 2015-09-30
Kwu/python docs

kxwu:


index.rst
```diff
@@ -0,0 +1,29 @@
+.. ld_python_client documentation master file, created by
```
jkodumal: The .rst files are generated, correct? I don't think we want to check them in.

kxwu: I actually think they're not automatically generated. I had to make them using a separate command, and RTD says that it looks for a conf.py and any .rst files to build, rather than building them itself.

jkodumal:
Obsolete.

PR #27, opened 2015-10-07
Add additional requirements to manifest

pstreule:
Version 0.19.0 is currently broken without this change.


jkodumal:
Thanks for this! Our integration tests were pointing directly to a SHA, so we weren't adequately testing the v0.19.0 artifact.

PR #32, opened 2015-11-15
Support New Relic Annotations

albertyw:


jkodumal:
I set up NewRelic on our python integration test wrapper. I don't see any errors, but I'm not seeing custom attributes appearing on the transactions.

albertyw:
I changed the direct call to `annotate_transaction` into a decorator so that it would catch every case `toggle` is called, plus I added an extra integration test.  I believe it should hit NewRelic in your test wrapper now.

albertyw:
Should there be a Config option for whether NR is enabled?  Currently, the client will send data to NR if it's detected to be installed, but I can think of some contrived examples where a user has NR installed but wants to disable setting NR custom params.  I can also port this ability to the ruby-client.

jkodumal:
Hmm. Tried again today, and not seeing the data annotated in newrelic.

albertyw:
Not sure what the problem is then.  The integration tests are using `toggle` and not some other function right?  Also, the integration tests are submitting transactions to new relic right?  I believe the `add_custom_parameter` only works when the transaction is submitted.  How are the integration tests run?  I don't understand how `integration-harness` works.

Issue #35, opened 2015-12-29
Strict requirements are no fun


crempp:
Would it be possible to update the requirements.txt to not use strict (==) requirements? This has forced me to downgrade requests to 2.4.0.

jkodumal:
Certainly-- we can make that change.

witten:
Additionally, it's generally considered bad practice (especially for a library) to slurp in requirements.txt and use it for a setup.py's install_requires. Check out this article on the subject: https://caremad.io/2013/07/setup-vs-requirement/

pkaeding:
From my understanding, it looks like the objections raised in that article are around making sure that you do not have strict requirements in the setup.py, right?

So unless I'm misinterpreting something, it should be safe to relax the requirements in requirements.txt, and still reference that from setup.py.

witten:
That's certainly part of it, but the other part is that libraries specifically shouldn't include a requirements.txt at all, because they shouldn't have any strict/concrete requirements (and concrete requirements is the purpose of a requirements file). So it really would be more in the spirit of that article not to have a requirements.txt for your library, and put all abstract/non-strict requirements directly in your setup.py instead.

PR #36, opened 2015-12-29
Switch to using compatible requirements everywhere

jkodumal:


jkodumal:
Addresses #35

albertyw:
The current version of requests is 2.9.1 (http://docs.python-requests.org/en/latest/) and `requests~=2.4.0` will only allow versions `2.4.*`.  It would require more testing, but perhaps `requests>=2.4.0` is the way to go?

albertyw:
On second thought, until pytest is testing different package versions along with different python versions, it would also be a good idea to set a max version so that a backwards incompatible release doesn't break the client.  `requests` and `future` are pretty darn stable, but I'm not sure about `sseclient` and `CacheControl`.

albertyw:
looks good to me

PR #47, opened 2016-06-30
Add polling + indirect support for streaming

drichelson:
I tried to change as little as possible, but that didn't work. See comments in diff.
This is deployed to restwrappers and harness is running: https://circleci.com/gh/launchdarkly/integration-harness/1126


ldclient/client.py
```diff
@@ -156,18 +108,19 @@ def __init__(self, api_key, config=None):
         self._offline = False
```
jkodumal: One thing we did in the other refactorings like this is change offline mode to a configuration parameter rather than a modifiable state. This makes it a lot easier to reason about things, and it makes it the case that we don't have to lazily initialize the streaming connection.

ldclient/client.py
```diff
@@ -35,36 +39,50 @@ def __init__(self,
                  verify=True,
                  defaults=None,
                  events=True,
-                 stream_processor_class=None,
-                 feature_store_class=None,
-                 feature_requester_class=None,
+                 update_processor_class=None,
```
drichelson: This model of having an update processor + a feature requester is consistent with the other SDKs

ldclient/client.py
```diff
@@ -35,36 +39,50 @@ def __init__(self,
...
         self.base_uri = base_uri.rstrip('\\')
+        self.get_latest_features_uri = self.base_uri + GET_LATEST_FEATURES_PATH
         self.events_uri = events_uri.rstrip('\\')
         self.stream_uri = stream_uri.rstrip('\\')
-        self.stream = stream
-        self.stream_processor_class = RequestsStreamProcessor if not stream_processor_class else stream_processor_class
-        self.feature_store_class = InMemoryFeatureStore if not feature_store_class else feature_store_class
+        self.stream_features_uri = self.stream_uri + STREAM_FEATURES_PATH
+
+        if update_processor_class:
```
drichelson: you can manually set the update processor class (ie twisted)- if so, the stream param is ignored.

ldclient/client.py
```diff
@@ -79,73 +97,7 @@ def default(cls):
         return cls()
 
 
-class InMemoryFeatureStore(FeatureStore):
```
drichelson: Moved to its own file

ldclient/twisted_impls.py
```diff
@@ -22,40 +22,33 @@ def __init__(self, api_key, config):
...
                 defer.returnValue(None)
 
         return run(True)
 
     @defer.inlineCallbacks
-    def _toggle(self, key):
+    def _get_all(self):
         hdrs = _headers(self._api_key)
-        uri = self._config.base_uri + '/api/eval/features/' + key
+        uri = self._config.get_latest_features_uri
```
drichelson: I moved all the uri paths to 'constants' in one file.

testing/test_integration.py
```diff
@@ -30,18 +30,10 @@ def fin():
     return server
 
 
-def test_toggle(server):
-    server.add_feature("foo", feature("foo", "jim")['foo'])
-    client = LDClient("apikey", Config(base_uri=server.url, events_uri=server.url))
-    wait_until(lambda: client.toggle("foo", user('xyz'), "blah") == "jim")
-
-
-def test_sse_init(server, stream):
```
drichelson: this is tested by test_toggle, and a bunch of times in the integration harness as well

testing/test_integration_twisted.py
```diff
@@ -29,14 +29,6 @@ def fin():
     request.addfinalizer(fin)
     return server
 
-
-@pytest.inlineCallbacks
-def test_toggle(server):
```
drichelson: this is tested by test_toggle, and a bunch of times in the integration harness as well

ldclient/twisted_redis.py
```diff
@@ -9,7 +9,14 @@ def create_redis_ldd_processor(api_key, config, store, **kwargs):
     return TwistedRedisLDDStreamProcessor(store, **kwargs)
 
 
-class TwistedRedisLDDStreamProcessor(StreamProcessor):
+class TwistedRedisLDDStreamProcessor(UpdateProcessor):
```
drichelson: Do we need this? It seems like this should be a RedisFeatureStore, not a Redis stream processor..

drichelson:
This is passing against restwrappers running locally, but is failing on Circle. I'm investigating, but I believe all this code to be correct.

ldclient/client.py
```diff
@@ -79,95 +100,33 @@ def default(cls):
...
-        try:
-            self._lock.rlock()
-            return self._initialized
-        finally:
-            self._lock.runlock()
-
-
 class LDClient(object):
-
     def __init__(self, api_key, config=None):
```
jkodumal: Can we add an option (as we have in go) to wait for initialization? That would get rid of the ugly `sleep` in thet demo code.

ldclient/twisted_redis.py
jkodumal: Indeed, I think this should be refactored into a RedisFeatureStore.

demo/demo.py
```diff
@@ -1,7 +1,25 @@
...
-    client = LDClient(apiKey)
+    apiKey = 'your api key'
+    config = Config(stream=False)
+    client = LDClient(apiKey, config)
+    user = {u'key': 'userKey'}
+    time.sleep(5)
+    print(client.toggle("update-app", user, False))
     print(client.api_key)
+
+    client._stop_consumers()
```
jkodumal: Wondering if you tested whether this will exit cleanly? On other platforms we've had issues with daemon threads so it's a good thing to check.

ldclient/client.py
```diff
@@ -35,41 +40,57 @@ def __init__(self,
...
-                 stream_processor_class=None,
-                 feature_store_class=None,
-                 feature_requester_class=None,
-                 consumer_class=None):
+                 update_processor_class=None,
+                 poll_interval=1,
+                 use_ldd=False,
+                 feature_store=InMemoryFeatureStore(),
+                 feature_requester_class=FeatureRequesterImpl,
+                 consumer_class=None,
```
jkodumal: Can we rename this to `event_consumer_class`?

ldclient/client.py
```diff
@@ -35,41 +40,57 @@ def __init__(self,
...
-        :type feature_requester_class: (str, Config) -> FeatureRequester
+        :type feature_requester_class: (str, Config, FeatureStore) -> FeatureRequester
         :param consumer_class: A factory for an EventConsumer implementation taking the event queue, api key, and config
         :type consumer_class: (queue.Queue, str, Config) -> EventConsumer
         """
         if defaults is None:
             defaults = {}
 
         self.base_uri = base_uri.rstrip('\\')
+        self.get_latest_features_uri = self.base_uri + GET_LATEST_FEATURES_PATH
```
jkodumal: This (and the stream uri) look to be concerns that are exclusive to their respective `UpdateProcessor` implementations. Can we move this logic there instead of leaving it in the general configuration?

ldclient/client.py
jkodumal: No changes, correct?

ldclient/client.py
jkodumal: In LDD mode there should not be an update processor-- check the Go code for the logic.

ldclient/polling.py
```diff
@@ -0,0 +1,34 @@
...
+                elapsed = time.time() - start_time
+                if elapsed < self._config.poll_interval:
+                    time.sleep(self._config.poll_interval - elapsed)
+
+    def initialized(self):
+        return self._running and self._store.initialized
+
+    def stop(self):
+        log.debug("Stopping PollingUpdateProcessor")
+        self._running = False
```
jkodumal: this won't stop anything-- you never check whether we should be running in the while loop

jkodumal:
I would recommend updating `version.py` now as part of this refactoring.

ldclient/polling.py
```diff
@@ -0,0 +1,34 @@
...
+
+class PollingUpdateProcessor(Thread, UpdateProcessor):
+    def __init__(self, api_key, config, requester, store):
+        Thread.__init__(self)
+        self.daemon = True
+        self._api_key = api_key
+        self._config = config
+        self._requester = requester
+        self._store = store
+        self._running = False
```
jkodumal: I don't *think* there are any implications of this, and I notice we use the same pattern in the `StreamProcessor`, but this doesn't look thread safe to me.

ldclient/requester.py
```diff
@@ -1,104 +1,44 @@
...
-        elif msg.event == 'delete':
-            key = payload['path'][1:]
-            # noinspection PyShadowingNames
-            version = payload['version']
-            store.delete(key, version)
-        else:
-            log.warning('Unhandled event in stream processor: ' + msg.event)
-
-
 class RequestsEventConsumer(Thread, EventConsumer):
```
jkodumal: Can we move this class out to its own file? It's a separate concern from feature flag requests, and it's more in line with the other implementations to push this into a different file.

ldclient/client.py
drichelson: These are also accessed in the twisted implementations, so I think we should leave them here.

ldclient/client.py
drichelson: correct

ldclient/polling.py
drichelson: The while loop will break out when self._running is set to False, but it won't interrupt a loop in progress.

jkodumal: Oh, of course. I missed that.

demo/demo.py
drichelson: Can you be more specific? do you mean does the app actually stop when you hit ctrl-c?

ldclient/polling.py
drichelson: Do you mean that the modification of self._running is not done under a lock?

demo/demo.py
jkodumal: Yes, exactly



		_____________________________

drichelson: well, handling something like ctrl-c is outside the scope of this library. I can tell you that calling this method will cleanly shut down the client.

jkodumal: The question is whether the client library creates any non-daemon threads that prevent applications from exiting cleanly. It is very much within the scope of the client library to ensure that consuming applications can shut down properly when using the library.

drichelson: We don't start any non-daemon threads. As mentioned, calling this method shuts down the client cleanly.

drichelson:
3 existing restwrappers: python with streaming, python polling, and twisted streaming:
https://circleci.com/gh/launchdarkly/integration-harness/1173 (python)
https://circleci.com/gh/launchdarkly/integration-harness/1174 (twisted)

ldclient/__init__.py
```diff
@@ -11,6 +13,27 @@
...
+"""Settings."""
+client = None
+api_key = None
+start_wait = 5
+config = Config()
+
+_lock = threading.Lock()
+
+
+def get():
```
drichelson: This follows the model used by segment to enforce the singleton pattern. See updated readme + restwrapper PR: https://github.com/launchdarkly/python-restwrapper/pull/1/files

ldclient/client.py
```diff
@@ -79,117 +93,81 @@ def default(cls):
...
+        self._store = self._config.feature_store
         """ :type: FeatureStore """
 
-        self._feature_requester = self._config.feature_requester_class(
-            api_key, self._config)
+        if self._config.offline:
+            log.info("Started LaunchDarkly Client in offline mode")
+            return
+
+        if self._config.events_enabled:
```
drichelson: Do we want to keep this option?

jkodumal: Yes, we should be adding it to other SDKs too, ideally.

ldclient/polling.py
jkodumal: Correct.

drichelson:
Added python-redis restwrapper, it should run with this build: https://circleci.com/gh/launchdarkly/integration-harness/1176

drichelson:
The megabuild: https://circleci.com/gh/launchdarkly/integration-harness/1176 failed the polling restwrapper, but I ran it again: https://circleci.com/gh/launchdarkly/integration-harness/1177 and it passed. 2/3 pass.

demo/demo.py
```diff
@@ -1,7 +1,25 @@
...
+ch.setLevel(logging.DEBUG)
+formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+ch.setFormatter(formatter)
+root.addHandler(ch)
+
 if __name__ == '__main__':
-    apiKey = 'feefifofum'
-    client = LDClient(apiKey)
+    api_key = 'api_key'
+    client = LDClient(api_key, start_wait=10)
```
pkaeding: Should this be calling `ldclient.get()` like the README says?

ldclient/__init__.py
```diff
@@ -11,6 +13,27 @@
...
+    try:
+        _lock.acquire()
```
drichelson: @pkaeding suggested looking at ways to not always acquire the lock as it could be heavily contended during normal use.

pkaeding:
As there are a number of breaking changes here (I noticed the major version bump, so that is good), we should be sure to call out those changes in the release notes.

ldclient/redis_feature_store.py
```diff
@@ -0,0 +1,114 @@
...
+    def __setitem__(self, key, value):
+        pass
+
+
+class RedisFeatureStore(FeatureStore):
+    def __init__(self,
+                 url='redis://localhost:6379/0',
+                 prefix='launchdarkly',
+                 max_connections=16,
+                 expiration=15,
```
pkaeding: these parameters should probably be documented

pkaeding:
lgtm
:+1:
looks good to me
merge away
I approve
canned replies are great.

jkodumal:
Let's have this PR target a `v2` branch, instead of master.

ldclient/__init__.py
```diff
@@ -11,6 +13,34 @@
...
+"""Settings."""
+client = None
+api_key = None
+start_wait = 5
+config = Config()
+
+_lock = ReadWriteLock()
+
+
+def get():
```
drichelson: Added double-check locking. Please run through how this can break.

jkodumal: This looks reasonable to me.

drichelson:
ok- I'm closing this PR and merging this branch into v2.

PR #52, opened 2016-08-08
Allow blank user key when evaluating flags. Fix redis all()

drichelson:
Allow blank user key when evaluating flags. Fix redis all()


jkodumal:
👍

PR #55, opened 2016-08-16
Add Twisted support for LDD mode

drichelson:
Part of this is recycled from before we removed Twisted support.


ldclient/feature_store.py
```diff
@@ -10,24 +10,24 @@ def __init__(self):
         self._initialized = False
         self._features = {}
 
-    def get(self, key):
+    def get(self, key, callback):
```
drichelson: Technically this is an interface change that should require us to change the version to 3.0..

jkodumal: Does this work in the non-twisted world?

drichelson: yes- Locally I ran the integration harness against the redis and in-memory feature store

ldclient/feature_store.py
```diff
@@ -10,24 +10,24 @@ def __init__(self):
...
             if f is None:
                 log.debug("Attempted to get missing feature: " + str(key) + " Returning None")
-                return None
+                return callback(None)
             if 'deleted' in f and f['deleted']:
                 log.debug("Attempted to get deleted feature: " + str(key) + " Returning None")
-                return None
-            return f
+                return callback(None)
+            return callback(f)
```
pkaeding: I'm probably missing something, but is this class used by both Twisted and non-Twisted? 

The interface docs below state 

> `:return: The feature value. None if not found`

So shouldn't this execute `callback(f)` but `return f`?

And if I'm understanding this correctly, then I think you can give the callback parameter a default value (of a no-op function) to lessen the impact of the API changes.

ldclient/interfaces.py
```diff
@@ -8,18 +8,18 @@ class FeatureStore(object):
...
-        :return: The feature flag data
-        :rtype: dict
+        :param callback: The function that accepts the feature data and returns the feature value
+        :type callback: function
+        :return: The feature value. None if not found
         """
 
     @abstractmethod
-    def all(self):
+    def all(self, callback):
```
pkaeding: the callback parameter should be documented

ldclient/twisted_client.py
```diff
@@ -0,0 +1,80 @@
...
+            value = self._evaluate_internal(flag, user)
+            value.addBoth(cb)
+            return value
+
+        return _get_off_variation(flag)
+
+    def _evaluate_internal(self, flag, user):
+        def check_prereq_results(result):
+            prereq_ok = True
+            for r in result:  # r is a tuple of 2 booleans: (error, prereqMatches)
```
pkaeding: Can we use [NamedTuples](https://docs.python.org/2/library/collections.html#collections.namedtuple) to increase the readability?

ldclient/twisted_event_consumer.py
```diff
@@ -0,0 +1,88 @@
...
+
+from ldclient.interfaces import EventConsumer
+from ldclient.util import _headers, log
+
+
+class TwistedEventConsumer(EventConsumer):
+
+    def __init__(self, queue, sdk_key, config):
+        self._queue = queue
+        """ @type: queue.Queue """
```
pkaeding: should this be `:type:`?

testing/test_feature_store.py
```diff
@@ -66,43 +66,43 @@ def test_initialized(self, store):
     def test_get_existing_feature(self, store):
         store = self.base_initialized_store(store)
         expected = self.make_feature('foo', 10)
-        assert store.get('foo') == expected
+        assert store.get('foo', lambda x: x) == expected
```
pkaeding: It might be clearer to define this identity function once, and reference it in all the places.

pkaeding:
lgtm
:+1:
looks good to me
merge away
I approve
canned replies are great.

jkodumal:
Is there a twisted RESTWrapper in place?

drichelson:
@jkodumal This is only for LDD mode. We don't have LDD running in our restwrapper environment, but the twisted restwrapper passes locally.

PR #58, opened 2016-10-21
Fix indirect streaming messages.

drichelson:
Fix indirect streaming messages. I ran the integration harness with these changes (including a big feature flag that triggers an indirect/put) and it passed.

### DIFF
--- a/ldclient/streaming.py
+++ b/ldclient/streaming.py
@@ -46,20 +46,20 @@ class StreamingUpdateProcessor(Thread, UpdateProcessor):
 
     @staticmethod
     def process_message(store, requester, msg, ready):
-        payload = json.loads(msg.data)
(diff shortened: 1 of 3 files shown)


ldclient/streaming.py
```diff
@@ -46,27 +46,28 @@ def initialized(self):
 
     @staticmethod
     def process_message(store, requester, msg, ready):
-        payload = json.loads(msg.data)
```
drichelson: The indirect message payloads are not json, but instead are a list of flag keys (or one flag key in the case of indirect/patch). We were failing here when attempting to parse the json.

jkodumal:
👍

pkaeding:
lgtm
:+1:
looks good to me
merge away
I approve
canned replies are great.

PR #62, opened 2016-11-02
Add backoff to stream reconnect

drichelson:
Initial changes to initialization. This attempts to address the fact that we aggressively retry streaming connections. There are other improvements to make, but this one is bite-sized and has other corrections as well.

We're also now properly reporting that the client is initialized or not.

### DIFF
--- a/ldclient/client.py
+++ b/ldclient/client.py
@@ -35,8 +35,8 @@ class Config(object):
     def __init__(self,
                  base_uri='https://app.launchdarkly.com',
                  events_uri='https://events.launchdarkly.com',
-                 connect_timeout=2,
-                 read_timeout=10,
+                 connect_timeout=10,
+                 read_timeout=15,
                  events_upload_max_batch_size=100,
                  events_max_pending=10000,
                  stream_uri='https://stream.launchdarkly.com',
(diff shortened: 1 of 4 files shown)


ldclient/client.py
```diff
@@ -35,8 +35,8 @@ class Config(object):
     def __init__(self,
                  base_uri='https://app.launchdarkly.com',
                  events_uri='https://events.launchdarkly.com',
-                 connect_timeout=2,
-                 read_timeout=10,
+                 connect_timeout=10,
```
drichelson: These were lower than what we've seen in the field (5 seconds just for dns lookup) so I'm upping them here.

ldclient/client.py
```diff
@@ -149,10 +149,11 @@ def __init__(self, sdk_key, config=None, start_wait=5):
         log.info("Waiting up to " + str(start_wait) + " seconds for LaunchDarkly client to initialize...")
         update_processor_ready.wait(start_wait)
 
-        if self._update_processor.initialized:
+        if self._update_processor.initialized() is True:
```
drichelson: This was not evaluating correctly.. it would indicate success even when the streaming processor had not been initialized.

ldclient/streaming.py
```diff
@@ -30,8 +33,17 @@ def run(self):
                 for msg in messages:
                     if not self._running:
                         break
-                    if self.process_message(self._store, self._requester, msg, self._ready) is True:
+                    message_ok = self.process_message(self._store, self._requester, msg, self._ready)
+                    if message_ok is True and self._ready.is_set() is False:
                         self._ready.set()
+            except HTTPError as e:
+                if e.response is not None and e.response.status_code is not None:
```
drichelson: We now stop the reconnect loop if a 4xx response code is seen.

drichelson: which begs the question- is it possible for a 404 error to be transient? due to a misconfigured/misbehaving CDN?

pkaeding:
This definitely looks good. It might also be worth not retrying at all when we get a status in the 4xx range, since retrying the same request should result in the same status.

PR #64, opened 2016-11-05
Allow SDK key/config to be changed after client has been initialized.

drichelson:
The reasoning behind this is that it makes the reset SDK Key feature much more managable when a dynamic configuration is in place. 

The intention is that this is an api-breaking change for anyone using the new singleton flow found in __init__.py, but anyone still using the client constructor should not have to make any code changes.


drichelson:
[ch138]

demo/demo.py
```diff
@@ -15,11 +15,11 @@
 root.addHandler(ch)
 
 if __name__ == '__main__':
-    ldclient.sdk_key = 'sdk_key'
     ldclient.start_wait = 10
-    client = ldclient.get()
+    ldclient.set_sdk_key('YOUR_SDK_KEY')
```
drichelson: Example of new usage pattern.

ldclient/__init__.py
```diff
@@ -12,38 +12,87 @@
...
+            old_client.close()
+    finally:
+        __config = config
+        __lock.unlock()
+
+
+# 2 use cases:
+# 1. Initial setup: sets the sdk key for the uninitialized client
+# 2. Allows on-the-fly changing of the sdk key. When this function is called after the client has been initialized
+#    the client will get re-initialized with the new sdk key. In order for this to work, the return value of
```
drichelson: This is the pitfall of this approach, but I don't see another way to do it. any ideas @pkaeding ?

ldclient/client.py
```diff
@@ -27,79 +25,10 @@
 from cachecontrol import CacheControl
 from threading import Lock
 
-GET_LATEST_FEATURES_PATH = '/sdk/latest-flags'
```
drichelson: config code now in its own file.

ldclient/config.py
```diff
@@ -0,0 +1,171 @@
+from ldclient.event_consumer import EventConsumerImpl
```
drichelson: The refactored Config class is as immutable as python allows..

ldclient/__init__.py
```diff
@@ -12,38 +12,87 @@
...
+            return __client
     finally:
-        _lock.runlock()
+        __lock.runlock()
 
     try:
-        global client
-        _lock.lock()
-        if not client:
+        global __client
```
pkaeding: I don't think this is needed, since you declared `global __client` above.

ldclient/__init__.py
```diff
@@ -12,38 +12,87 @@
...
+                old_client = __client
+                __config = new_config
+                __client = new_client
+                old_client.close()
+        finally:
+            __lock.unlock()
+
+
+# the return value should not be assigned.
 def get():
```
pkaeding: In the README, you give an example of:

    ldclient.get("your sdk key")

Should this be:

    def get(sdk_key=None):

ldclient/client.py
pkaeding: In order to make this change backwards compatible, I think you could alias the `Config` class in this package

ldclient/client.py
```diff
@@ -27,79 +25,10 @@
...
-        return default if key not in self.defaults else self.defaults[key]
-
-    @classmethod
-    def default(cls):
-        return cls()
-
 
 class LDClient(object):
-    def __init__(self, sdk_key, config=None, start_wait=5):
+    def __init__(self, config=None, start_wait=5):
```
pkaeding: To try to make this backwards-compatible, I think you can also allow the sdk_key to be passed to this ctor, and apply it to the config if it is passed in.

If there is some reason not to do that (that I'm missing), then the config argument should not be optional-- it is not valid to create a client with no SDK key.

ldclient/event_consumer.py
```diff
@@ -13,11 +13,10 @@
 
 
 class EventConsumerImpl(Thread, EventConsumer):
-    def __init__(self, event_queue, sdk_key, config):
+    def __init__(self, event_queue, config):
```
pkaeding: This (and below) is a little harder to rationalize the backwards-compatible way of overlaying an optional `sdk_key` argument on the config.

Maybe doing that, and marking it `[deprecated](https://wiki.python.org/moin/PythonDecoratorLibrary#Generating_Deprecation_Warnings)` would be the way to go?  Then we can remove it in the next major release.

ldclient/__init__.py
pkaeding: I don't think we should say that it should *never* be assigned; that seems a bit extreme. It is fine to get the client at the beginning of a request handler that needs to get a bunch of flags. You just don't want to keep the reference long-term.  

On the other hand, we believe the population of users who need to change the SDK key at runtime is rather small. So most people can disregard this warning altogether if they like.  I think we can re-word this to be less scary, while still explaining enough of the implications of different use cases.

pkaeding:
I think this approach makes sense. I would like to see if we can possibly do it in a backwards-compatible way, to avoid needing a major release, and make upgrading easier for our users.  I'm not totally sure if that is possible, though.

ldclient/__init__.py
drichelson: I'm almost inclined to avoid this wording, except when documenting how to change the config in an already-running app.

ldclient/__init__.py
```diff
@@ -12,38 +12,87 @@
...
+# 2 use cases:
+# 1. Initial setup: sets the sdk key for the uninitialized client
+# 2. Allows on-the-fly changing of the sdk key. When this function is called after the client has been initialized
+#    the client will get re-initialized with the new sdk key. In order for this to work, the return value of
+#    ldclient.get() should never be assigned
+def set_sdk_key(sdk_key):
+    global __config
+    global __client
+    global __lock
+    if sdk_key is __config.sdk_key:
```
drichelson: yes

ldclient/__init__.py
drichelson: apologies- readme is not up to date.

ldclient/client.py
drichelson: These changes are api-breaking, so we'll be on a new major release anyway. It's one less change our customers will make, but is arguably cruft. I guess the argument is that it's one less thing they will have to change when updating the client.

ldclient/client.py
drichelson: it is valid to create an offline client with no sdk key, or a client with a redis feature store (events of course will fail or can be disabled)

ldclient/event_consumer.py
drichelson: These changes necesitate a major release (due to changes in singleton init flow). Does this change your take on this?

ldclient/client.py
drichelson: Also, directly creating a client is no longer the suggested route, so I prefer to have it only take in the config object.

pkaeding: Ahh, good point about the offline client. I get that we don't want to encourage people to directly create the client, but it might ease the transition if they could drop in the new version, and have everything work as it did before, modulo some deprecation warnings that they can then work to eliminate.

ldclient/__init__.py
```diff
@@ -12,38 +12,87 @@
...
-            client = LDClient(sdk_key, config, start_wait)
-        return client
+            __client = LDClient(__config, start_wait)
+        return __client
     finally:
-        _lock.unlock()
+        __lock.unlock()
+
+
+def init():
```
drichelson: My thinking here is that this could be called in server init code to pre-warm the client so the first request doesn't incur the penalty of connecting to the stream etc... on second thought I have removed it. Thoughts?

ldclient/client.py
```diff
@@ -27,9 +27,19 @@
 
 
 class LDClient(object):
-    def __init__(self, config=None, start_wait=5):
+    def __init__(self, sdk_key=None, config=None, start_wait=5):
         check_uwsgi()
-        self._config = config or Config.default()
+
+        if config is not None and sdk_key is not None:
```
pkaeding: I think this should check if the skd_key is set in both places, not if any config is passed in. Some existing clients might be using a modified config (changed timeouts, etc), but passing in the `sdk_key` separately.

drichelson: good point- this could break existing patterns..

pkaeding:
lgtm
:+1:
looks good to me
merge away
I approve
canned replies are great.

jkodumal:
👍

PR #68, opened 2017-01-10
Fix redis feature store so it returns default when redis is unavailable.

drichelson:
Fix redis feature store so it returns default when redis is unavailable.


jkodumal:
👍

PR #81, opened 2018-04-16
Fix import of parse_requirements

### LINKED ISSUE #80: Install broken on pip 10.x
From the [pip CHANGELOG](https://pip.pypa.io/en/stable/news/#b1-2018-03-31):

> Move all of pip's APIs into the pip._internal package, properly reflecting the fact that pip does not currently have any public APIs. (#4696, #4700)

```
Collecting ldclient-py==4.0.6 (from -r requirements.txt (line 74))
  Downloading ldclient-py-4.0.6.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-qsq8HS/ldclient-py/setup.py", line 9, in <module>
        from pip.req import parse_requirements
    ImportError: No module named req

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-qsq8HS/ldclient-py/
```

mark-adams:
Starting in pip 10.x, parse_requirements was moved to an internal
namespace to help demonstrate the fact that it is not a public API.

As a workaround until a broader decision can be made about how this
should look long term, this change attepts to perform the import using
the pip 10.x path and if it fails, falls back on the older path.

Fixes #80.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,10 @@ except ImportError:
 import sys
 import uuid
 
-from pip.req import parse_requirements
+try:
+    from pip._internal.req import parse_requirements
+except ImportError:
+    from pip.req import parse_requirements
 
 ldclient_version='5.0.3'
 


eli-darkly:
We're going to go with [a different approach](https://github.com/launchdarkly/python-client/pull/82) just to avoid importing any internal APIs, but thanks for submitting this.

mark-adams:
No worries. Just trying to get something out quick and fast since this caused a lot of pain for a lot of people when pip 10.x was released. :+1:

PR #96, opened 2018-09-04
Stop using Exception.message

mattbriancon:
`Exception.message` has been deprecated since Python 2.6 in favor of leveraging `Exception.__str__()` or `Exception.args`.


eli-darkly:
Good catch, thanks.

eli-darkly:
I'll add a unit test to verify that exception logging works correctly in all versions of Python.

eli-darkly:
Version 6.4.1 includes this fix.

mattbriancon:
Thanks for the quick turn around!

PR #99, opened 2018-09-24
Handle None return from feature store .all()

thieman:
Fixes a bug where `self._store.all()` returns `None` without throwing an Exception. This would go on to immediately fail on line 328 as `None` does not have `.items()`. Instead, this makes us go through the established error handler and bail out with `FeatureFlagsState(False)`

cc @mattbriancon

### DIFF
--- a/ldclient/client.py
+++ b/ldclient/client.py
@@ -321,6 +321,8 @@ class LDClient(object):
         with_reasons = kwargs.get('with_reasons', False)
         try:
             flags_map = self._store.all(FEATURES, lambda x: x)
+            if flags_map is None:
+                raise ValueError("flags_map is None, aborting")
         except Exception as e:
             log.error("Unable to read flags for all_flag_state: %s" % e)
             return FeatureFlagsState(False)
--- a/ldclient/client.py
+++ b/ldclient/client.py
@@ -322,7 +322,7 @@ class LDClient(object):
         try:
             flags_map = self._store.all(FEATURES, lambda x: x)
             if flags_map is None:
-                raise ValueError("flags_map is None, aborting")
+                raise ValueError("feature store error")
         except Exception as e:
             log.error("Unable to read flags for all_flag_state: %s" % e)
             return FeatureFlagsState(False)


eli-darkly:
Hmm. I agree that the current implementation is a problem, but I'm not sure if this is the best place to fix it - especially if that means having to raise and catch another exception. The other option would be to just change the feature store so it returns an empty dict, rather than None, for these error conditions. That would be consistent with the behavior of the single-item `get` method, where if there is a Redis error it will just behave the same as if it could not retrieve the item (though it will log the error).

eli-darkly:
On the other hand... your approach will result in setting the `valid` property of the state object to False, which is appropriate since this is an error condition. So, OK. Thanks for catching this!

eli-darkly:
Actually I have one request: could you change the exception message from "flags_map is None, aborting" to "feature store error"? That would be more descriptive - no one should need to know this particular variable name.

thieman:
Potentially dumb question: is the condition on this line actually an error / invalid state? Is there a different valid state the store is expected to be in if it's initialized but empty? https://github.com/launchdarkly/python-client/blob/master/ldclient/redis_feature_store.py#L72

eli-darkly:
> is the condition on this line actually an error / invalid state?

Well, this is more confusing than I thought.

I was thinking that `hgetall` should return an empty dict, rather than `None` or `""` in that case. My assumption was that the hash key would still exist even if there are no items. However, looking again at how we've implemented `init` (lines 47-51), it seems that where there are no items the hash key will not exist.

Unfortunately, the redis-py documentation says nothing at all about how `hgetall` is supposed to behave in this case. So I took a look at the source code and it's still not entirely clear to me, but [this line](https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L408) suggests that it should return an empty dict. I will verify that experimentally.

eli-darkly:
Indeed, `hgetall` always returns a dict even if the key does not exist, assuming there wasn't some other kind of Redis error. So I would say that yes, if `flags_map` is None or "" then that is an error condition.

Issue #115, opened 2019-03-29
setting user attributes to non-strings when strings are expected can cause problems


eli-darkly:
We've identified some potential problems that can happen if any of the non-custom user attributes that are expected to be strings (that is, all of them except `:anonymous`) are instead set to a number or some other type. The Python SDK, like the other SDKs in weakly-typed languages, does not prevent this since the user object is simply a dictionary.

1. LaunchDarkly will not process analytics events for such a user.
2. If you set `"secondary"` to a non-string, and try to evaluate a flag that has a percentage rollout, the evaluation will fail.

We believe this behavior is the same in all existing versions of the Python SDK.

Our planned fix is to make the SDK coerce these values to strings as needed. To avoid causing unexpected changes in behavior for existing code, we will not enforce strict typing in this major version.

PR #116, opened 2019-03-29
prepare 6.8.4 release

eli-darkly:
### Fixed:
- Setting user attributes to non-string values when a string was expected would cause analytics events not to be processed. Also, in the case of the "secondary" attribute, this could cause evaluations to fail for a flag with a percentage rollout. The SDK will now convert attribute values to strings as needed. ([#115](https://github.com/launchdarkly/python-client/issues/115))
- If `track` or `identify` is called without a user, the SDK now logs a warning, and does not send an analytics event to LaunchDarkly (since it would not be processed without a user).


ldclient/flag.py
```diff
@@ -10,13 +10,18 @@
...
 from ldclient.versioned_data_kind import FEATURES, SEGMENTS
 
 __LONG_SCALE__ = float(0xFFFFFFFFFFFFFFF)
 
 __BUILTINS__ = ["key", "ip", "country", "email",
                 "firstName", "lastName", "avatar", "name", "anonymous"]
 
+__USER_ATTRS_TO_STRINGIFY_FOR_EVALUATION__ = [ "key", "secondary" ]
+# Currently we are not stringifying the rest of the built-in attributes prior to evaluation, only for events.
+# This is because it could affect evaluation results for existing users (ch35206).
```
torchhound: Should we expose a clubhouse ticket number?

eli-darkly: I don't see why not. It's not sensitive information and it may be helpful to us later.

torchhound: Fair enough, just checking.

PR #119, opened 2019-04-08
Error closing client when LaunchDarkly relay proxy in daemon mode

benpagedelphix:
Do not attempt to read the `_update_processor` attribute when closing the `client` if the client is running in `ldd` mode because initialization of the `_update_processor` attribute does not occur if `use_ldd` is set to true in the config.

### DIFF
--- a/ldclient/client.py
+++ b/ldclient/client.py
@@ -157,6 +157,8 @@ class LDClient(object):
             return
         if self._event_processor:
             self._event_processor.stop()
+        if self._config.use_ldd:
+            return
         if self._update_processor and self._update_processor.is_alive():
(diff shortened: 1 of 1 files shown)


eli-darkly:
This definitely needs to be fixed, but I think we're going to go in a slightly different direction: making sure that all of the object's attributes are always set at initialization time, so we don't need to go "check this other attribute first because if it has a certain value, then the attribute I want won't exist." In other words, the `close` method should not need to know the details of the `use_ldd` behavior.

benpagedelphix:
That makes sense to me, thanks for fixing this ASAP

PR #121, opened 2019-04-24
compare sdk key value, not identity

jpgimenez:
the SDK key must be compared by value, using `==`, not by identity, using `is`, because the important thing here is that the value didn't change, comparing by identity cause false reinitializations of the client

### DIFF
--- a/ldclient/__init__.py
+++ b/ldclient/__init__.py
@@ -68,7 +68,7 @@ def set_sdk_key(sdk_key):
     sdk_key_changed = False
     try:
         __lock.rlock()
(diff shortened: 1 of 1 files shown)


eli-darkly:
Thanks for catching this!

jpgimenez:
thanks for merging!

eli-darkly:
This was a pretty bad bug, so we'll try to get a patch release with the fix out ASAP.

eli-darkly:
The fix is now in the 6.9.1 release.

PR #130, opened 2019-10-08
Allow explicitly proxying only ld requests

gangeli:
This commit adds a configuration parameter allowing the configuration of
a proxy that should be used when connecting to LaunchDarkly's API. This
allows only this service to be proxied, without having to specify the
global `http_proxy` environment variable.

Our use-case for this was calling LaunchDarkly from within an AWS VPC without external internet access, where we need a proxy to talk to any service not in our VPC but don't want to have the proxy the traffic within the VPC. Explicitly setting a proxy configuration for just LaunchDarkly seemed to me the cleanest solution to this.

### DIFF
--- a/ldclient/config.py
+++ b/ldclient/config.py
@@ -43,7 +43,8 @@ class Config(object):
                  offline=False,
                  user_keys_capacity=1000,
                  user_keys_flush_interval=300,
-                 inline_users_in_events=False):
+                 inline_users_in_events=False,
+                 http_proxy=None):
         """
         :param string sdk_key: The SDK key for your LaunchDarkly account.
         :param string base_uri: The base URL for the LaunchDarkly server. Most users should use the default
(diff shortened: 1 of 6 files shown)


eli-darkly:
Thanks for putting this together. We've filed it in our backlog for review. In the meantime, can you verify by manual testing that this works as intended? Automated tests would also be a good idea but would be trickier to do— probably would require setting up an embedded HTTP server in the tests— so we might have to take care of that part, but it'd be good to know if it at least passes a smoke test.

gangeli:
I've tested it locally and confirmed that I'm hitting the configured proxy. We have not yet deployed it into a live environment, but that should happen in the next couple of days. So, smoke test passes but we'll know for sure once I push it to staging :)

eli-darkly:
I'm curious about one thing: you mentioned setting `http_proxy`. Was that a typo for `https_proxy`, or were you able (before implementing this change) to cause the LaunchDarkly SDK to use a proxy just by setting `http_proxy`?

eli-darkly:
On further investigation, it looks to me like the ability to specify a proxy via the `https_proxy` environment variable was broken— this may have happened when we changed which HTTP library we were using. The specific logic you added looks fine, but since I may need to add an extra fix, I'm going to accept it to a different branch and work on it from there. Thanks!

eli-darkly:
We did not include this in the latest release because we had to put out a bugfix release (also related to proxies) first, but it will be in 6.11.0.

eli-darkly:
This is now available in the 6.11.0 release.

gangeli:
[Belated] thanks!

PR #133, opened 2019-11-20
Addresses https://bugs.python.org/issue34850

bunchesofdonald:
**Requirements**

**Related issues**
Python 3.8 made it so that comparisons done via `is` to a literal (str, int, etc.) now throw a SyntaxWarning: https://bugs.python.org/issue34850 In our test suite we convert warnings to errors so this is causing a fairly major issue for us and upgrading to py3.8.

Nothing has changed about the code itself I've just changed all the places where `is` was used to do a comparison to a literal to use `==` instead.

No alternatives were considered, this is the only way to fix this particular issue.

**Additional context**

Add any other context about the pull request here.

### DIFF
--- a/ldclient/config.py
+++ b/ldclient/config.py
@@ -290,5 +290,5 @@ class Config(object):
         return self.__http_proxy
 
     def _validate(self):
-        if self.offline is False and self.sdk_key is None or self.sdk_key is '':
+        if self.offline is False and self.sdk_key is None or self.sdk_key == '':
             log.warning("Missing or blank sdk_key.")
--- a/ldclient/fixed_thread_pool.py
+++ b/ldclient/fixed_thread_pool.py
@@ -63,7 +63,7 @@ class FixedThreadPool(object):
     def _run_worker(self):
         while True:
             item = self._job_queue.get(block = True)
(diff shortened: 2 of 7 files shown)


bunchesofdonald:
That py3.3 failure seems to be that typing 3.7.4.1 dropped support for python 3.3: ```ERROR: You need Python 2.7 or 3.4+ to install the typing package.```

eli-darkly:
Thanks for catching this... and we will add a 3.8 CI build, as we should've done earlier.

The 3.3 issue you mentioned is known to us and is somewhat confusing. Since the latest `typing` is clearly not compatible with 3.3, I'm not sure why it is trying to install that version as a transitive dependency.

eli-darkly:
Oh - I do know what the 3.3 problem is. It's pulling in `typing` as a transitive dependency of `expiringdict`, which started using it in a recent release. We just need to pin `expiringdict`.

eli-darkly:
We've released 6.11.1 with this fix, as well as the fix for 3.3. Thanks again.

bunchesofdonald:
That's great, thank you so much!

PR #145, opened 2020-09-15
Allow authenticating with proxy

gangeli:
This commit allows for authenticating with a proxy configured with the
`http_proxy` environment variable. Authentication requires passing a
header, and is not parsed by urllib3 from the proxy_url.

This has been tested and is working in our own environment, but no additional tests have been run.

**Requirements**

- [ ] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)
- [ ] I have validated my changes against all supported platform versions

**Related issues**

Square uses a proxy with authentication for its internal services, and this is unable to connect through that proxy.

The URL for the proxy is parsed to explicitly parse out authentication and pass it in as a header.

N/A

**Additional context**

N/A

### DIFF
--- a/ldclient/impl/http.py
+++ b/ldclient/impl/http.py
@@ -54,11 +54,18 @@ class HTTPFactory(object):
                 ca_certs=ca_certs
                 )
         else:
+            # Get proxy authentication, if provided
+            url = urllib3.util.parse_url(proxy_url)
+            proxy_headers = None
+            if url.auth != None:
+                proxy_headers = urllib3.util.make_headers(proxy_basic_auth=url.auth)
+            # Create a proxied connection
             return urllib3.ProxyManager(
                 proxy_url,
                 num_pools=num_pools,
                 cert_reqs=cert_reqs,
-                ca_certs = ca_certs
+                ca_certs = ca_certs,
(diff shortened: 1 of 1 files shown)


eli-darkly:
This looks reasonable, but an automated test would be good to have rather than only manual testing. That's one of the PR requirements mentioned on that checklist, and we know that some kinds of changes are impractical to write unit tests for but I don't think this is one of them— you'll see in `test_feature_requester.py` there are some tests for the proxy URL setting, and I think it'd be fairly straightforward to extend those to include a case where the URL has auth parameters and we verify that the HTTP request contains the expected header.

gangeli:
Would it be possible to offload testing onto you all? I'm happy to test manually to make sure it works in our environment, but all things equal would prefer to avoid spelunking through the testing code. In particular, I'm unsure how to test that authentication is actually being set with your mocked server.

eli-darkly:
OK. I'll merge the PR into another branch so I can add the tests and verify things on our end before it goes to master. Thanks for the contribution.

eli-darkly (approved):

eli-darkly:
This fix is in the 6.13.2 release. Thanks again!

PR #147, opened 2020-10-27
prepare 7.0.0 release

LaunchDarklyCI:
## [7.0.0] - 2020-10-27
This major release is for Python compatibility updates and removal of deprecated APIs. It introduces no new functionality except type hints.

### Added:
- Added [type hints](https://docs.python.org/3/library/typing.html) to all SDK methods. Python by itself does not enforce these, but commonly used development tools can provide static checking to trigger warnings or errors if the wrong type is used.

### Changed:
- Python 2.7, 3.3, and 3.4 are no longer supported. The minimum Python version is now 3.5.
- The first parameter to the `Config` constructor, `sdk_key`, is now required. Previously it was possible to omit the `sdk_key` from the `Config` and specify it separately when initializing the SDK. Now, it is always in the `Config`.

### Removed:
- Removed `ldclient.set_sdk_key()`. The correct way to do this now, if you are using the singleton client method `ldclient.get()`, is to call `ldclient.set_config()` with a `Config` object that contains the SDK key.
- Removed the optional SDK key parameter from the [`LDClient`](https://launchdarkly-python-sdk.readthedocs.io/en/latest/api-main.html#ldclient.client.LDClient) constructor. You must now provide a configuration parameter of type [`Config`](https://launchdarkly-python-sdk.readthedocs.io/en/latest/api-main.html#ldclient.config.Config), and set the SDK key within the `Config` constructor: `LDClient(Config(sdk_key = "my-sdk-key", [any other config options]))`. Previously, it was possible to specify the SDK key as a single string parameter and omit the `Config` object—`LDClient("my-sdk-key")`—although this would cause a deprecation warning to be logged; specifying both a key and a `Config` was always an error.
- Removed the individual HTTP-related parameters such as `connect_timeout` from the [`Config`](https://launchdarkly-python-sdk.readthedocs.io/en/latest/api-main.html#ldclient.config.Config) type. The correct way to set these now is with the [`HTTPConfig`](https://launchdarkly-python-sdk.readthedocs.io/en/latest/api-main.html#ldclient.config.HTTPConfig) sub-configuration object: `Config(sdk_key = "my-sdk-key", http = HTTPConfig(connect_timeout = 10))`.
- Removed all other types, parameters, and methods that were deprecated as of the last 6.x release.


torchhound (approved):

bwoskow-ld (changes requested):
There are character encoding issues in a few places in the changelog. Also there are unresolved merge conflicts.

bwoskow-ld (approved):

PR #148, opened 2020-11-20
Fix check uwsgi options

andrefreitas:
Even if uwsgi is avaiable in sys.modules, you need to check if uwsgi.opt is available. This code was broken in latest version of uwsgi when running a process without uwsgi. The uwsgi documentation unfortunately is not clear about this https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html

Also the link for LaunchDarkly documentation is broken.

### DIFF
--- a/ldclient/util.py
+++ b/ldclient/util.py
@@ -34,13 +34,16 @@ def check_uwsgi():
     if 'uwsgi' in sys.modules:
         # noinspection PyPackageRequirements,PyUnresolvedReferences
         import uwsgi
+        if not hasattr(uwsgi, 'opt'):
+            # means that we are not running under uwsgi
+            return
 
         if uwsgi.opt.get('enable-threads'):
             return
         if uwsgi.opt.get('threads') is not None and int(uwsgi.opt.get('threads')) > 1:
             return
         log.error("The LaunchDarkly client requires the 'enable-threads' or 'threads' option be passed to uWSGI. "
-                    'To learn more, see http://docs.launchdarkly.com/v1.0/docs/python-sdk-reference#configuring-uwsgi')
+                    'To learn more, see https://docs.launchdarkly.com/sdk/server-side/python#configuring-uwsgi')
 
 
 class Event:


eli-darkly:
Can you say a bit more about how you discovered this behavior? The uwsgi documentation indeed doesn't seem to say anything at all about a circumstance in which uwsgi would be present but uwsgi.opt wouldn't, and I haven't found anything in the uwsgi release notes to indicate when (or why) this behavior changed. I don't see any harm in adding this extra check, since we don't want to throw an error due to uwsgi.opt being missing if it is missing, but I'd like to understand what this would actually mean.

ldclient/util.py
```diff
@@ -34,13 +34,15 @@ def check_uwsgi():
...
+            # means that we are not running under uwsgi
+            return
 
         if uwsgi.opt.get('enable-threads'):
             return
         if uwsgi.opt.get('threads') is not None and int(uwsgi.opt.get('threads')) > 1:
             return
-        log.error("The LaunchDarkly client requires the 'enable-threads' or 'threads' option be passed to uWSGI. "
-                    'To learn more, see http://docs.launchdarkly.com/v1.0/docs/python-sdk-reference#configuring-uwsgi')
+        log.error("The LaunchDarkly client requires the 'enable-threads' or 'threads' option be passed to uWSGI. ")
```
eli-darkly, line 45: Rather than removing the link, let's just change it to the correct URL: https://docs.launchdarkly.com/sdk/server-side/python#configuring-uwsgi

andrefreitas: Thanks @eli-darkly, I updated the documentation link 👍

andrefreitas:
@eli-darkly thanks for your prompt feedback. I also don't understand from uWSGI documentation unfortunately. We detected this issue running LaunchDarkly under RQ.

For the sake of checking if this is the "normal" behaviour, I created a simple http server that runs with and without uWSGI and tried to use the ld client and import uWSGI, and when it doesn't run under uWSGI the import fails as the documentation say: https://github.com/andrefreitas/test-uwsgi

```
flask_1  | [2020-11-25 15:43:30,852] ERROR in app: Exception on / [GET]
flask_1  | Traceback (most recent call last):
flask_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
flask_1  |     response = self.full_dispatch_request()
flask_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
flask_1  |     rv = self.handle_user_exception(e)
flask_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
flask_1  |     reraise(exc_type, exc_value, tb)
flask_1  |   File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
flask_1  |     raise value
flask_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
flask_1  |     rv = self.dispatch_request()
flask_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
flask_1  |     return self.view_functions[rule.endpoint](**req.view_args)
flask_1  |   File "/app/server.py", line 12, in index
flask_1  |     import uwsgi # fails in flask
flask_1  | ModuleNotFoundError: No module named 'uwsgi'
```

I don't like not understanding why rq changes this, but I also see that adding this extra check doesn't produce any harm, so I am fine with the tradeoff of not understanding fully at this moment the root cause. 

So are you fine in proceeding like this?

eli-darkly:
@andrefreitas Yes, I agree that there's no harm in the extra check. I was just hoping to be able to write a release note that was a little more specific about what circumstances this might apply to, but I guess the note will have to be somewhat vague.

eli-darkly (approved):
Thanks for catching and addressing these issues.

eli-darkly:
We've released version 7.0.1 which includes this fix.

andrefreitas:
@eli-darkly me and the Beyond Pricing engineering team appreciate your collaboration. Thanks 🙏

PR #169, opened 2022-04-21
Add support for extra Redis connection params

danie1k:
**Requirements**

- [x] I have added test coverage for new or changed functionality
- [x] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)
- [x] I have validated my changes against all supported platform versions

**Related issues**

\-

**Describe the solution you've provided**

The `redis.ConnectionPool.from_url()` can accept various number of kwargs.  
Among them, very imporatnt one is the `decode_components=True` which must be enabled in case redis password contains special characters.  
After [rfc1738](https://www.ietf.org/rfc/rfc1738.txt): 

> In addition, octets may be encoded by a character triplet consisting of the character "%" followed by the two hexadecimal digits (from "0123456789ABCDEF") which forming the hexadecimal value of the octet.
>
> ...
> Unsafe:
> Characters can be unsafe for a number of reasons. (...) All unsafe characters must always be encoded within a URL.

In redis-py version < 4, if the flag `decode_components` is set to `False`, then the escaped characters stays escaped and connection to the Redis server cannot be established:

https://github.com/redis/redis-py/blob/2c9f41f46991f94f0626598600d1023d4e12f0bc/redis/connection.py#L947-L950

**Describe alternatives you've considered**

There is none.

**Additional context**

\-

PR #170, opened 2022-04-28
Add support for extra Redis connection params

danie1k:
Replaces #169 

----

**Requirements**

**Related issues**

\-

The `redis.ConnectionPool.from_url()` can accept various number of kwargs.  
Among them, very imporatnt one is the `decode_components=True` which must be enabled in case redis password contains special characters.  
After [rfc1738](https://www.ietf.org/rfc/rfc1738.txt): 

> In addition, octets may be encoded by a character triplet consisting of the character "%" followed by the two hexadecimal digits (from "0123456789ABCDEF") which forming the hexadecimal value of the octet.
>
> ...
> Unsafe:
> Characters can be unsafe for a number of reasons. (...) All unsafe characters must always be encoded within a URL.

In redis-py version < 4, if the flag `decode_components` is set to `False`, then the escaped characters stays escaped and connection to the Redis server cannot be established:

https://github.com/redis/redis-py/blob/2c9f41f46991f94f0626598600d1023d4e12f0bc/redis/connection.py#L947-L950

There is none.

**Additional context**

\-

### DIFF
--- a/ldclient/integrations/__init__.py
+++ b/ldclient/integrations/__init__.py
@@ -144,7 +144,8 @@ class Redis:
     def new_feature_store(url: str='redis://localhost:6379/0',
                           prefix: str='launchdarkly',
                           max_connections: int=16,
-                          caching: CacheConfig=CacheConfig.default()) -> CachingStoreWrapper:
+                          caching: CacheConfig=CacheConfig.default(),
+                          **conn_options: Any) -> CachingStoreWrapper:
         """
         Creates a Redis-backed implementation of :class:`~ldclient.interfaces.FeatureStore`.
         For more details about how and why you can use a persistent feature store, see the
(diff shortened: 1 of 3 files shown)


ldclient/integrations/__init__.py
```diff
@@ -144,7 +144,8 @@ class Redis:
     def new_feature_store(url: str='redis://localhost:6379/0',
                           prefix: str='launchdarkly',
                           max_connections: int=16,
-                          caching: CacheConfig=CacheConfig.default()) -> CachingStoreWrapper:
+                          caching: CacheConfig=CacheConfig.default(),
+                          **conn_options: Any) -> CachingStoreWrapper:
```
keelerm84, line 148: Having a single catch-all for a single purpose limits our ability to add more optional parameters in the future, unless they are co-mingled with the conn_options.

It would be preferable to us if you could add a named parameter `redis_opts` which served to hold those pass through configuration options. For now, we would pull those values from max_connections or `redis_opts` if it is set there. We can clean this up in our next major release, but for now, this keeps our ability to make future changes more flexible.

keelerm84 (changes requested):
Thank you for your contribution! I appreciate you taking the time to put together this PR. I had one small change and then I think this looks good.

keelerm84:
Thanks again for your contribution. At this point, I'm going to merge this into a contrib branch, pull it into our private repo, make the tweaks I suggested, and then I'll work on getting this released. Thank you again for the time and effort you put into this PR!

PR #172, opened 2022-05-04
Fix missing py.typed

### LINKED ISSUE #171: py.typed file is missing
issue https://github.com/launchdarkly/python-server-sdk/issues/165 was closed via https://github.com/launchdarkly/python-server-sdk/pull/166

unfortunately that wasn't a complete fix, as the `py.typed` file that was added to the repo was not included in the source distribution sent to pypi: https://files.pythonhosted.org/packages/55/0a/7bf239d8550745fdde6d7a26788c1e9aac2a1628586e22cac8a1d8d61bae/launchdarkly-server-sdk-7.4.1.tar.gz

...so mypy still gives:
```
error: Skipping analyzing "ldclient": module is installed, but missing library stubs or py.typed marker
```

anentropic:
alternatively add to `package_data` in `setup.py` https://stackoverflow.com/a/63172998/202168

Fixes #171

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,4 @@ include test-requirements.txt
 include consul-requirements.txt
 include dynamodb-requirements.txt
 include redis-requirements.txt
+include ldclient/py.typed


keelerm84:
Thank you for your contribution. 

When I was testing this locally, it seemed I also needed to include include_package_data=True into the setup.py file. Did you run into this?

anentropic:
I haven't tested it TBH, you may be right

But I believe `include_package_data=True` would only be needed if the file was being included via `package_data` rather than `MANIFEST.in`, I think they are either/or methods

keelerm84:
According to [this site](https://setuptools.pypa.io/en/latest/userguide/datafiles.html), I think we need to include it:

> Setuptools offers three ways to specify this most common type of data files to be included in your package’s [1](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#datafiles). First, you can simply use the include_package_data keyword, e.g.:
> 
> from setuptools import setup, find_packages
> setup(
>     ...
>     include_package_data=True
> )
> This tells setuptools to install any data files it finds in your packages. The data files must be specified via the [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) file.

anentropic:
Thanks!  Is there somewhere I can watch or get notified when this is in a released version?

eli-darkly:
@anentropic We generally add another comment on an accepted PR after we've released those changes, but also, you can use GitHub's "Watch" on the repository and it will notify you when there's a release, then unwatch it if you don't want to follow every release after that.

PR #177, opened 2022-06-15
prepare 7.5.0 release

LaunchDarklyReleaseBot:
### Changed:
- Removed upper version restriction on expiringdict. This was originally necessary to allow compatibility with older Python versions which are no longer supported.


eli-darkly (changes requested):
I've made an edit to the changelog text for clarity, but I also have another question I'd like to discuss internally.

PR #193, opened 2023-01-31
prepare 8.1.0 release

LaunchDarklyReleaseBot:
### Added:
- Introduced support for an `application` config property which sets application metadata that may be used in LaunchDarkly analytics or other product features. This does not affect feature flag evaluations.


kinyoklion (changes requested):
Double period in the change description.

keelerm84:
> Double period in the change description.

Fix in GH UI and in releaser. Will fix previous changelog entry in 7.x. I would do it here but I'm trying to avoid future merge conflicts.

kinyoklion (approved):

PR #196, opened 2023-02-07
prepare 8.1.1 release

LaunchDarklyReleaseBot:
### Fixed:
- Fixed indexing error raised by calling `all_flags_state` while using the `TestData` data source.


ldclient/version.py
```diff
@@ -1 +1 @@
-VERSION = "8.1.0"
+VERSION = "7.6.1"
```
louis-launchdarkly, line 1: Something seems to be wrong with this release - as the title of this PR doesn't match the version number here.

louis-launchdarkly (changes requested):
Is there any issue with this release PR?

ldclient/version.py
keelerm84: That's because I released something on the 7.x branch and then had to merge that forward to the main branch.

Releaser doesn't actually change the version as part of this PR. It will merge this in first, then change it in another commit when it also adds the changelog. See [here][1] and [here][2].

[1]: https://github.com/launchdarkly/python-server-sdk/commit/db883f5b0b76598e17d5154d490a806dd2a53b5a
[2]: https://github.com/launchdarkly/python-server-sdk/commit/ce5dd0526f048c65168b585a0a9be5975f591056

louis-launchdarkly (approved):

ldclient/version.py
louis-launchdarkly: Thank you for explaining! I just approved the change.

PR #219, opened 2023-08-30
Stream and polling processor logging updates to warn on recoverable errors

fritzdj:
We are using a logging integration for internal alerting and we want to be able to filter out certain error logs where the LD client is running into a "recoverable" error if possible. Today we noticed there were HTTP responses with a 500 status being returned when StreamingUpdateProcessor was running -- this looked to be a minor blip. Could we log these at the warning level instead of the error level if these are 'expected' blips?

### DIFF
--- a/ldclient/impl/datasource/polling.py
+++ b/ldclient/impl/datasource/polling.py
@@ -38,8 +38,11 @@ class PollingUpdateProcessor(UpdateProcessor):
                 log.info("PollingUpdateProcessor initialized ok")
                 self._ready.set()
         except UnsuccessfulResponseException as e:
-            log.error(http_error_message(e.status, "polling request"))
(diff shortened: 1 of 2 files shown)


keelerm84:
@fritzdj thank you for this contribution. These changes look good to me. Can you also update the behavior in the polling processor so the behavior stays consistent between the two data sources?

Thank you!

fritzdj:
> @fritzdj thank you for this contribution. These changes look good to me. Can you also update the behavior in the polling processor so the behavior stays consistent between the two data sources?
> 
> Thank you!

@keelerm84, thanks for the review! I just pushed the same change for `PollingUpdateProcessor`.

keelerm84:
Fixed in 8.1.6

And once again, thank you for your contribution!

PR #235, opened 2023-12-13
fix(deps): Bump certifi to fix CVE-2023-37920 and CVE-2022-23491

keelerm84:


requirements.txt
```diff
@@ -1,4 +1,4 @@
-certifi>=2018.4.16
+certifi>=2023.07.22
```
keelerm84, line 1: Bumping this is probably considered a major version change since applications that are pinned to a lesser dependency can't actually use this next version without changing their dependencies as well.

cwaldren-ld (approved):

PR #246, opened 2024-01-09
feat: Redact anonymous attributes within feature events

keelerm84:


ldclient/impl/events/event_context_formatter.py
```diff
@@ -17,24 +17,25 @@ def __init__(self, all_attributes_private: bool, private_attributes: List[str]):
             if ar.valid:
                 self._private_attributes.append(ar)
 
-    def format_context(self, context: Context) -> dict:
```
cwaldren-ld, line 20: Same comment on other PRs re: adding a named function, not disturbing existing

cwaldren-ld (approved):

ldclient/impl/events/event_context_formatter.py
keelerm84: Also done.

PR #256, opened 2024-01-23
build: adding provenance generation to release-please workflow

rsoberano-ld:
**Requirements**

- [ ] I have followed the repository's [pull request submission guidelines](../blob/main/CONTRIBUTING.md#submitting-pull-requests)
- [ ] I have validated my changes against all supported platform versions

**Related issues**

Provide links to any issues in this repository or elsewhere relating to this pull request.

Using Github SLSA generator to generate build provenance for `python-server-sdk` 

Implementation based off of previous SDK SLSA integrations with release-please and Python-specific guidance here: https://sethmlarson.dev/python-and-slsa#generating-a-provenance-attestation

Provide a clear and concise description of any alternative solutions or features you've considered.

**Additional context**

Add any other context about the pull request here.

### DIFF
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -50,6 +50,4 @@ jobs:
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.7.0
     with:
       base64-subjects: "${{ needs.build-publish.outputs.package-hashes }}"
-      upload-assets: true
-      upload-tag-name: TBD
-        
+      upload-assets: ${{ !inputs.dry_run }}
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -14,6 +14,8 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    outputs:
+      package-hashes: ${{ steps.build.outputs.package-hashes}}
     steps:
       - uses: actions/checkout@v4
 
@@ -31,8 +33,23 @@ jobs:
           ssm_parameter_pairs: '/production/common/releasing/pypi/token = PYPI_AUTH_TOKEN'
 
       - uses: ./.github/actions/build
+        id: build
 
       - uses: ./.github/actions/publish
         with:
           token: ${{env.PYPI_AUTH_TOKEN}}
           dry_run: ${{ inputs.dry_run }}
+
+  release-provenance:
+    needs: [ 'build-publish' ]
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      id-token: write
+      contents: write
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.7.0
+    with:
+      base64-subjects: "${{ needs.build-publish.outputs.package-hashes }}"
+      upload-assets: true
+      upload-tag-name: TBD
+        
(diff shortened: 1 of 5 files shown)


PROVENANCE.md
```diff
@@ -0,0 +1,28 @@
...
+$ curl --location -O \
+  https://github.com/launchdarkly/python-server-sdk/releases/download/VERSION/multiple.intoto.jsonl
+
+# Run slsa-verifier to verify provenance against package artifacts 
+$ slsa-verifier verify-artifact \
+--provenance-path multiple-provenance.intoto.jsonl \
+--source-uri github.com/launchdarkly/launchdarkly-server-sdk \
+launchdarkly_server_sdk-VERSION-py3-none-any.whl
+
+TBD OUTPUT
```
rsoberano-ld, line 23: TBD pending test output - @keelerm84 what's the best way to test out the updates to the release workflow?

rsoberano-ld:
@keelerm84 does the `manual-publish.yml` workflow publish anything to Github releases?

.github/workflows/manual-publish.yml
```diff
@@ -31,8 +33,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      id-token: write
+      contents: write
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.7.0
+    with:
+      base64-subjects: "${{ needs.build-publish.outputs.package-hashes }}"
+      upload-assets: true
+      upload-tag-name: TBD
```
keelerm84, line 54: TBD should be fixed here.

.github/workflows/manual-publish.yml
```diff
@@ -31,8 +33,23 @@ jobs:
...
+    needs: [ 'build-publish' ]
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      id-token: write
+      contents: write
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.7.0
+    with:
+      base64-subjects: "${{ needs.build-publish.outputs.package-hashes }}"
+      upload-assets: true
```
keelerm84, line 53: For this workflow, you could tie this into the dry_run dispatch input. That way you can verify the hashes but you don't upload anything?

PROVENANCE.md
keelerm84: So there isn't a good way to test this without actually publishing something.

You can run the manual-publish with the dry run option (as I've mentioned in another comment), but actually going through the whole process isn't something that is easily done.

You can use the internal release please testing repo and just modify this code to not actually publish anything as well. I've done that a lot during initial dev.

rsoberano-ld: Leaving the example output empty for now until we're able to publish something (so it'll come in a separate commit after).

Let's coordinate on doing a manual-publish dry run next week to at least test most of the functionality here, and we can figure out a path for a full test after, though it sounds like the only way to make sure this works all the way through is to do this on an actual release. At the very least, the provenance generation happens last and won't fail the rest of the release steps

.github/workflows/manual-publish.yml
rsoberano-ld: Since we're triggering manual publishes in the case that a `release-please` workflow fails, the provenance generator should by default use the tag associated with the branch HEAD that we're building off of, so I think we can safely leave that parameter off here.

keelerm84:
> @keelerm84 does the `manual-publish.yml` workflow publish anything to Github releases?

It probably uploads a compressed copy of the source code, but the actual publication of the package is handled through pypi.

keelerm84 (approved):

PR #261, opened 2024-01-30
chore: Enable server-side-polling in contract tests (#260)

keelerm84:


keelerm84:
@cwaldren-ld backporting this from the main branch.

cwaldren-ld (approved):

PR #262, opened 2024-01-30
fix: Gzip header should only be set on polling

keelerm84:


keelerm84:
When I originally made this change, it passed testing against our live endpoints. However, it doesn't seem like upstream is actually honoring these settings after all. If they do start supporting this encoding, then it looks like this would fail for streaming.

So for now, I'm only going to enable it on the portion I know should have it implemented.

cwaldren-ld (approved):

PR #301, opened 2024-08-07
feat: `NO_PROXY` environment variable can be used to override `HTTP(S)_PROXY` values

keelerm84:
When determining if a proxy should be used, the SDK would:

1. Check the `config.http_config.http_proxy` value. If that is set, use
   that value without further consideration.
2. If the target URI is `https`, use the value from the `HTTPS_PROXY`
   environment variable.
3. If the target is `http`, use `HTTP_PROXY` instead.

The SDK will now support another environment variable -- `NO_PROXY`.
This variable can be set to a comma-separated list of hosts to exclude
from proxy support, or the special case '*' meaning to ignore all hosts.

The `NO_PROXY` variable will only take affect if the SDK isn't
explicitly configured to use a proxy as specified in #1 above.

### DIFF
--- a/ldclient/testing/impl/test_http.py
+++ b/ldclient/testing/impl/test_http.py
@@ -1,0 +1,57 @@
+import pytest
+
+from typing import Optional
+from ldclient.impl.http import _get_proxy_url
+
+
+@pytest.mark.parametrize(
+    'target_uri, no_proxy, expected',
+    [
+        ('https://secure.example.com', '', 'https://secure.proxy:1234'),
+        ('http://insecure.example.com', '', 'http://insecure.proxy:6789'),
+
+        ('https://secure.example.com', 'secure.example.com', None),
+        ('https://secure.example.com', 'secure.example.com:443', None),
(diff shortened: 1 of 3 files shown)


ldclient/testing/impl/test_http.py
```diff
@@ -0,0 +1,36 @@
...
+        ('https://secure.example.com', 'example.com:80', 'https://secure.proxy:1234'),
+
+        ('http://insecure.example.com', 'insecure.example.com', None),
+        ('http://insecure.example.com', 'insecure.example.com:443', 'http://insecure.proxy:6789'),
+        ('http://insecure.example.com', 'insecure.example.com:80', None),
+        ('http://insecure.example.com', 'wrong.example.com', 'http://insecure.proxy:6789'),
+
+        ('http://insecure.example.com', 'example.com', None),
+        ('http://insecure.example.com', 'example.com:443', 'http://insecure.proxy:6789'),
+        ('http://insecure.example.com', 'example.com:80', None),
```
keelerm84, line 26: I need to do some additional work to support the case where the target is missing the scheme, or it has a custom port specified.

louis-launchdarkly (approved):

PR #302, opened 2024-08-07
chore: Fix test warnings

keelerm84:


ldclient/testing/impl/test_attribute_ref.py
```diff
@@ -38,7 +38,7 @@ def test_ref_simple_with_leading_slash(self, input: str, unescaped: str):
         assert a.depth == 1
         assert a[0] == unescaped
 
-    @pytest.mark.parametrize("input", [])
+    @pytest.mark.parametrize("input", [("name"), ("name/with/slashes"), ("name~0~1with-what-looks-like-escape-sequences")])
```
keelerm84, line 41: This test was never testing anything, which is pretty silly.

ldclient/testing/test_ldclient.py
```diff
@@ -75,14 +75,12 @@ def test_client_has_null_update_processor_in_ldd_mode():
         assert isinstance(client._update_processor, NullUpdateProcessor)
 
 
-@pytest.mark.skip("Can't currently use a live stream processor in tests because its error logging will disrupt other tests.")
```
keelerm84, line 78: And this is no longer true it seems, so I'm removing those two skips.

louis-launchdarkly (approved):

PR #319, opened 2024-11-21
chore: Fix broken pypi publish action

keelerm84:
According to [this GitHub issue][1], the `pypa/gh-action-pypi-publish`
action cannot be used in nested actions. This commit inlines the usage
of this action into the top-level workflows instead.

[1]: https://github.com/pypa/gh-action-pypi-publish/issues/299


cwaldren-ld (approved):
Some tests are failing

.github/workflows/manual-publish.yml
```diff
@@ -35,10 +35,11 @@ jobs:
       - uses: ./.github/actions/build
         id: build
 
-      - uses: ./.github/actions/publish
+      - name: Publish package distributions to PyPI
+        if: ${{ inputs.dry_run == false }}
+        uses: pypa/gh-action-pypi-publish@release/v1
```
webknjaz, line 40: FYI, the security recommendation is to have this in a dedicated job and avoid giving elevated OIDC privileges to transitive build dependencies.

.github/workflows/manual-publish.yml
```diff
@@ -35,10 +35,11 @@ jobs:
...
         with:
-          token: ${{env.PYPI_AUTH_TOKEN}}
-          dry_run: ${{ inputs.dry_run }}
+          password: ${{env.PYPI_AUTH_TOKEN}}
```
webknjaz, line 42: This shouldn't be necessary with OIDC: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

PR #347, opened 2025-08-19
chore: Refactor FDv1 behind the datasystem interface

keelerm84:


ldclient/impl/datasystem/fdv1.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        update_processor = self._make_update_processor(
+            self._config, self._store_wrapper, set_on_ready
+        )
+        self._update_processor = update_processor
+        update_processor.start()
+
+    def stop(self):
+        if self._update_processor is not None:
+            self._update_processor.stop()
```
jsonbailey, line 110: I don't fully know the logic here, but since we are setting the `_update_processor` in the start method should we set it back to None after it stops? Would there ever be a case where you could call the start method on a previously stopped processor? It probably doesn't matter in this code since the start method will just overwrite it.

keelerm84: I don't think it will matter. As you point out, calling start again would override it. But no one is going to be interacting with the datasystem directly. That's all internal to the SDK so shouldn't be an issue.

PR #373, opened 2025-11-14
chore: Separate status check from synchronizer functionality

keelerm84:
In the previous setup, we would only check the fallback or recovery
conditions once the synchronizer returned an update. If the synchronizer
was stuck, or nothing was changing in the environment, we would never
check the conditions.

This configuration also exposed an interesting behavior. If the
synchronizer cannot connect, it will emit error updates. Each time we
receive an error, we check if we have failed to initialize for the last
10 seconds. If so, we re-create the primary synchronizer.

When it continues to fail, the first update will trigger the condition
check. And since it has still failed for 10 seconds, it will immediately
error out. With this change, we can be assured a synchronizer is given
at least 10 seconds to try before the condition is evaluated.

---

> [!NOTE]
> Separates FDv2 fallback/recovery checks from synchronizer emissions via a timer+reader thread, fixes initial status timestamp, removes streaming context manager hooks, and skips flaky file-source tests in CI.
> 
> - **FDv2 core**:
>   - Periodically evaluates conditions independent of updates: `_consume_synchronizer_results` now uses a queue, background reader thread, and `RepeatingTask` timer to trigger regular status checks and act without waiting for sync updates.
>   - Minor flow tweak when starting the secondary synchronizer.
> - **Status handling**:
>   - Initialize `DataSourceStatus.since` with `time.time()` in `DataSourceStatusProviderImpl`.
> - **Streaming**:
>   - Remove context-manager methods (`__enter__`, `__exit__`) from `StreamingDataSource`.
> - **Tests**:
>   - Update `test_two_phase_init` to ensure version progression and add readiness/change events.
>   - Mark file data source tests as flaky-skipped when `LD_SKIP_FLAKY_TESTS` is set.
> - **CI**:
>   - Set `LD_SKIP_FLAKY_TESTS: true` for test steps on Linux and Windows jobs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fa34244b05b8df82f97a8ed39bf1026ca8d2f527. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ldclient/testing/impl/datasystem/test_fdv2_datasystem.py
+++ b/ldclient/testing/impl/datasystem/test_fdv2_datasystem.py
@@ -21,7 +21,7 @@ def test_two_phase_init():
     # Set this to true, and then to false to ensure the version number exceeded
     # the initializer version number. Otherwise, they start as the same version
     # and the latest value is ignored.
-    td_initializer.update(td_initializer.flag("feature-flag").on(True))
+    td_synchronizer.update(td_initializer.flag("feature-flag").on(True))
     td_synchronizer.update(td_synchronizer.flag("feature-flag").on(False))
     data_system_config = DataSystemConfig(
         initializers=[td_initializer.build_initializer],


ldclient/impl/datasystem/fdv2.py
```diff
@@ -457,17 +496,14 @@ def _consume_synchronizer_results(
...
-                    return False, False
-
         except Exception as e:
             log.error("Error consuming synchronizer results: %s", e)
             return True, False
         finally:
             synchronizer.stop()
+            timer.stop()
+
+            sync_reader.join(0.5)
```
jsonbailey: What is the purpose of the half second timeout here? My understanding is that if it doesn't stop in .5 seconds, execution will move on and that thread will still continue to process.

jsonbailey: Not a blocker, just curious.

PR #381, opened 2025-11-25
feat: adding file data source as an intializer

joker23:
**Requirements**

This PR will:
1. add the file data source to the custom fdv2 datasystem configuration
2. modify the initialization logic so it aligns with spec. Namely, we will continue through the initializer list until we either run out of intializers (in which case the synchronizers will try to do a valid initialization) OR if an intializer fetched a proper basis with a selector.

---

> [!NOTE]
> Introduces a file-based initializer and updates FDv2 to mark ready only when an initializer provides a selector, with tests validating initializer ordering and early completion.
> 
> - **Core**:
>   - Add `file_ds_builder(paths)` returning `_FileDataSourceV2` initializer in `ldclient/datasystem.py`; import `_FileDataSourceV2`.
> - **FDv2 Initialization Logic** (`ldclient/impl/datasystem/fdv2.py`):
>   - After applying an initializer basis, set ready only if `basis.change_set.selector` is defined; then return to proceed to synchronizers.
> - **Tests** (`ldclient/testing/impl/datasystem/test_fdv2_datasystem.py`):
>   - Add tests ensuring initializers run until first success and stop on first initializer that provides a selector; verify transition to synchronizers.
>   - Include file-based data source usage via `file_ds_builder` with temp JSON files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 04a2c538e5d0f1f87f35feaaa949ae55d3cb8716. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ldclient/impl/datasystem/fdv2.py
+++ b/ldclient/impl/datasystem/fdv2.py
@@ -410,8 +410,7 @@ class FDv2(DataSystem):
                 self._store.apply(basis.change_set, basis.persist)
 
                 # Set ready event if an only if a selector is defined for the changeset
-                selector_is_defined = basis.change_set.selector is not None and basis.change_set.selector.is_defined()
-                if selector_is_defined and not set_on_ready.is_set():
+                if basis.change_set.selector is not None and basis.change_set.selector.is_defined():
                     set_on_ready.set()
                     return
             except Exception as e:
--- a/ldclient/testing/impl/datasystem/test_fdv2_datasystem.py
+++ b/ldclient/testing/impl/datasystem/test_fdv2_datasystem.py
@@ -437,11 +437,10 @@ def test_fdv2_stays_on_fdv1_after_fallback():
     assert flag is not None
 
 
-def test_fdv2_with_file_to_polling_initializers():
+def test_fdv2_initializer_should_run_until_success():
     """
-    Test that FDv2 can be initialized with a file data source and a polling data source.
-    In this case the results from the file data source should be overwritten by the
-    results from the polling datasource.
+    Test that FDv2 initializers will run in order until a successful run. Then
+    the datasystem is expected to transition to run synchronizers.
     """
     initial_flag_data = '''
 {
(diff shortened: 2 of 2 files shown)


ldclient/impl/datasystem/fdv2.py
```diff
@@ -409,9 +409,11 @@ def _run_initializers(self, set_on_ready: Event):
                 # Apply the basis to the store
                 self._store.apply(basis.change_set, basis.persist)
 
-                # Set ready event
-                if not set_on_ready.is_set():
+                # Set ready event if an only if a selector is defined for the changeset
+                selector_is_defined = basis.change_set is not None and basis.change_set.selector.is_defined()
+                if selector_is_defined and not set_on_ready.is_set():
                     set_on_ready.set()
```
joker23: Would like some thoughts on the ready set... given the new conditions for when things are ready, I wonder if we need to test for `not set_on_ready.is_set()` here. The reasoning is that, the intializer process will only ever set this once and return now

ldclient/testing/impl/datasystem/test_fdv2_datasystem.py
```diff
@@ -432,3 +435,112 @@ def test_fdv2_stays_on_fdv1_after_fallback():
     store = fdv2.store
     flag = store.get(FEATURES, "fdv1-flag", lambda x: x)
     assert flag is not None
+
+
+def test_fdv2_with_file_to_polling_initializers():
```
keelerm84: These tests reference polling, but they are using the test data sources instead. So we should probably update the language.

ldclient/testing/impl/datasystem/test_fdv2_datasystem.py
```diff
@@ -432,3 +435,112 @@ def test_fdv2_stays_on_fdv1_after_fallback():
...
+
+        def listener(_: FlagChange):
+            nonlocal count
+            count += 1
+
+        fdv2.flag_tracker.add_listener(listener)
+
+        fdv2.start(set_on_ready)
+        assert set_on_ready.wait(1), "Data system did not become ready in time"
+        assert count == 2, "Invalid initializer process"
```
keelerm84: Are we sure there isn't a timing issue in this test? You are testing for 2 callbacks, but it may actually do 3 -- one for each initialization, and then another when the sync runs the first time.

joker23: Ah I think I misunderstood these test datasource... I thought that the test datasources will only give a response/trigger a callback if we call the `update` function? eg https://github.com/launchdarkly/python-server-sdk/pull/381/files/064f65c76135c385a6fdcb0209caa0e7a8cf872c#diff-1cee15fc62f5439ab90479c81c5cf48672134012d9b81254303605674e9c4fb4R467 ...if not then what would be a good way to test this? just wait for the synchronizer to finish and make it count 3?

joker23:
@keelerm84 I changed up the tests a bit and I think they shouldn't run into race conditions anymore

PR #390, opened 2025-12-15
feat: Allow not sending evaluation event when evaluating a flag

pranjalv123:
**Requirements**

We'd like to be able to evaluate a flag without sending an evaluation event to LaunchDarkly. This is useful, for example, for internal tooling that helps debug issues by listing the value of a flag for a particular context. We don't want to send an evaluation event to LD because it doesn't constitute a "real" evaluation so we don't want to consider it as evidence that a flag is still being evaluated and shouldn't be cleaned up.

- we can get this behavior by evaluating `all_flags_state`, but that's a much more expensive operation, and doesn't allow us to pass custom contexts on a per-flag basis
- we can do this through the LD API, but that's slower and is rate limited.

---

> [!NOTE]
> Adds a per-call `send_events` flag to `variation`/`variation_detail` to evaluate without emitting events, with internal wiring and a test.
> 
> - **Client API**:
>   - Add `send_events` optional parameter (default `True`) to `LDClient.variation` and `LDClient.variation_detail` to allow evaluating flags without emitting analytics events.
>   - Thread `send_events` through `_evaluate_internal`, gating all event emissions (`new_unknown_flag_event`, evaluation result events, and default events).
>   - Update method docstrings accordingly.
> - **Tests**:
>   - Add `test_no_event_when_send_events_false` verifying no events are produced when `send_events=False`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c9d92a7f978e3ec780f3c30680c4d693a1158d9b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ldclient/testing/test_ldclient_events.py
+++ b/ldclient/testing/test_ldclient_events.py
@@ -318,3 +318,12 @@ def test_no_event_for_existing_feature_with_invalid_context():
         bad_context = Context.create('')
         assert 'default' == client.variation('feature.key', bad_context, default='default')
         assert count_events(client) == 0
+
+
+def test_no_event_when_send_events_false():
+    feature = build_off_flag_with_value('feature.key', 'value').track_events(True).build()
+    store = InMemoryFeatureStore()
+    store.init({FEATURES: {feature.key: feature.to_json_dict()}})
+    with make_client(store) as client:
+        assert 'value' == client.variation(feature.key, context, default='default', send_events=False)
+        assert count_events(client) == 0


jsonbailey:
@pranjalv123 can you share a little more about your use case? Are you wanting to have this available in your production app where some flag evaluations send events and others do not, or for an internal app where you can evaluate all flags without sending events?

We do support disabling events via the config (send_events) but it is a global option and I'm unsure if that would meet your needs.

pranjalv123:
> Are you wanting to have this available in your production app where some flag evaluations send events and others do not?

Yes, it's this. We primarily use this for internal tooling/analytics, but because we want to calculate the launchdarkly context exactly the same way as we do in our production app, it's easiest to do the evaluation from our production app. I can imagine a world where we initialize a separate client to do the non-tracking evaluations, but that would have some overhead in terms of redis connections (for relay proxy in daemon mode), memory usage, complexity, etc.

PR #450, opened 2026-06-25
chore: Extract shared sans-I/O cores and add config read-protocols

jsonbailey:
## What

Two pieces of async-SDK foundation, laid on the sync side and **behavior-identical to `main`**:

**1. Shared sans-I/O `_common` modules.** Pulls I/O-free logic out of the sync data paths into `impl/client_common.py`, `impl/datasystem/fdv2_common.py`, `impl/events/event_processor_common.py` — imported by both the sync code and the future async siblings. `client.py` / `event_processor.py` / `fdv2.py` route through them; `fdv2.py` re-exports the moved names via `__all__` so existing import paths still resolve.

**2. Config read-protocols + `build()` widening.** Introduces structural `Protocol`s in `config.py` describing the config surface each consumer reads, so a future `AsyncConfig` can satisfy the same contracts without a fork:
- `SdkIdentityConfig` — `sdk_key`, `wrapper_name`, `wrapper_version`, `application`; parameter type of `get_environment_metadata` / `secure_mode_hash`.
- `DataSourceBuilderConfig(SdkIdentityConfig)` — adds `base_uri`, `stream_base_uri`, `http`, `initial_reconnect_delay`, `poll_interval`, `payload_filter_key`, `_instance_id`; parameter type of `DataSourceBuilder.build()`.
- `PrivateAttributesConfig` — `all_attributes_private`, `private_attributes`; parameter type of `EventOutputFormatter`.

`Config` explicitly inherits `DataSourceBuilderConfig` + `PrivateAttributesConfig`. `DataSourceBuilder.build()` and the `datasourcev2` data sources are widened from `Config` to `DataSourceBuilderConfig`. Fully backward-compatible: `Config` satisfies all three protocols, so every existing caller works unchanged.

## Why

Foundation for the async Python SDK (epic SDK-60). The `_common` modules + config protocols let the sync and async clients share identical I/O-free behavior and config contracts instead of duplicating them. The async siblings and `AsyncConfig` land in separate slices.

## Parity

Behavior-identical to `main`: the `_common` extraction moves code verbatim, and the protocols/`build()` widening are types-only (`Config` satisfies the protocols; `build()` still receives a `Config` at runtime). Worth a careful sync-behavior review.

## Validation

- `uv run mypy ldclient` — clean
- `uv run isort --check ldclient contract-tests` / `uv run pycodestyle ldclient contract-tests` — clean
- `uv run pytest ldclient/testing --ignore=integrations` — 935 passed

Refs SDK-2602 (epic SDK-60).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,6 +78,10 @@ A special case is the module `ldclient.impl`, and any modules within it. Everyth
 
 So, if there is a class whose existence is entirely an implementation detail, it should be in `impl`. Similarly, classes that are _not_ in `impl` must not expose any public members (i.e. symbols that do not have an underscore prefix) that are not meant to be part of the supported public API. This is important because of our guarantee of backward compatibility for all public APIs within a major version: we want to be able to change our implementation details to suit the needs of the code, without worrying about breaking a customer's code. Due to how the language works, we can't actually prevent an application developer from referencing those classes in their code, but this convention makes it clear that such use is discouraged and unsupported.
 
+### Sync/async parity
+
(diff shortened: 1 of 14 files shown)


ldclient/impl/client_core.py
```diff
@@ -0,0 +1,66 @@
...
+    ApplicationMetadata,
+    EnvironmentMetadata,
+    SdkMetadata
+)
+from ldclient.version import VERSION
+
+
+def get_environment_metadata(config: Config) -> EnvironmentMetadata:
+    sdk_metadata = SdkMetadata(
+        name="python-server-sdk",
```
keelerm84: Do we need to make a distinction for async here?

keelerm84 (approved):

ldclient/impl/client_core.py
jsonbailey: I've made this composable with the client controlling the value and will add an -async suffix to the async client.

keelerm84 (approved):

PR #464, opened 2026-07-21
feat: Add async FDv1 streaming and data source status tracking

jsonbailey:
## Overview

Adds the async FDv1 streaming data source and its companion async data source status tracking, plus a new shared sans-I/O `datasource_common` module. This is one slice of the async Python SDK work.

## What's included

- **Async FDv1 streaming data source** (`ldclient/impl/datasource/async_streaming.py`) — driven by `ld_eventsource`'s `AsyncSSEClient` and owning its own `aiohttp` session via `make_client_session`. It consumes SSE `put`/`patch`/`delete` events and pushes the resulting data into the async data source update sink.

- **Async data source status tracking** (`ldclient/impl/datasource/async_status.py` — `AsyncDataSourceUpdateSinkImpl` and friends). The streaming source pushes data and status transitions into this update sink, so the two are runtime-coupled and land together in this slice.

- **New shared sans-I/O `datasource_common` module** (`ldclient/impl/datasource/datasource_common.py`) — `STREAM_ALL_PATH`, `sink_or_store`, and `parse_path`. Both the sync data sources and the new async streaming source use it.

- **Behavior-preserving extraction of the sync data sources.** `streaming.py` and `polling.py` are routed through `datasource_common`: the inline `_sink_or_store`/`_parse_path`/`STREAM_ALL_PATH`/`ParsedPath` definitions are removed in favor of importing the shared helpers, with call sites switched accordingly. No behavior change. The shared module is also needed by the upcoming async polling slice.

- **Unified status provider (sync/async).** The async status provider was unified with the sync one: the redundant `AsyncDataSourceStatusProviderImpl` (which had no async-specific behavior — byte-identical to the sync provider apart from its sink type) is removed, and the async data system now reuses the sync `DataSourceStatusProviderImpl`. To let both sinks satisfy it structurally, the sync provider's `update_sink` parameter was widened to a small status-only `Protocol` (`_DataSourceStatusSource`). This is the same interface-segregation pattern as the PR 1b config read-protocols and mirrors the already-unified big-segment status provider. This adds a small behavior-preserving change to sync `status.py`.

## Notes

- The async public class carries the experimental `.. caution::` block.
- No CHANGELOG entries or version bumps (handled at release time).

Tracked internally: SDK-2743

---

> [!NOTE]
> **Medium Risk**
> Touches core flag data ingestion (streaming, store versioning, change notifications) for the new async path; sync changes are mostly refactors to shared helpers plus a widened status-provider type.
> 
> **Overview**
> Adds **async FDv1 streaming** (`AsyncStreamingUpdateProcessor`) and **`AsyncDataSourceUpdateSinkImpl`** so SSE `put`/`patch`/`delete` updates flow through the async feature store with status and flag-change notifications, mirroring the sync stack.
> 
> Introduces **`datasource_common`** (`STREAM_ALL_PATH`, `sink_or_store`, `parse_path`, polling endpoint constant) and wires **sync** streaming/polling/feature requester through it instead of duplicated helpers—no intended behavior change on the sync path.
> 
> **Async store contract:** `AsyncFeatureStore.delete` now returns whether the delete was applied; the async sink only updates dependency tracking and flag listeners when upsert/delete actually write, avoiding spurious events on stale versions.
> 
> **Status provider:** `DataSourceStatusProviderImpl` accepts a small `_DataSourceStatusSource` protocol so one provider works with sync and async sinks.
> 
> Async streaming owns an **aiohttp session** built lazily in `_run` via `make_client_session` when no SSE factory is injected, with teardown on stop/errors; extensive tests cover streaming, status sink, and session lifecycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a32fc16791cc9e4620b91ef7bebbe1b94412b8c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/impl/datasource/async_streaming.py
+++ b/ldclient/impl/datasource/async_streaming.py
@@ -1,24 +1,24 @@
+"""
+Default implementation of the streaming component.
+"""
+
+# currently excluded from documentation - see docs/README.md
+
 import json
 import time
-from threading import Thread
-from typing import Optional
+from typing import Any, Optional
 from urllib import parse
 
-from ld_eventsource import SSEClient
-from ld_eventsource.actions import Event, Fault
-from ld_eventsource.config import (
-    ConnectStrategy,
-    ErrorStrategy,
-    RetryDelayStrategy
-)
+from ld_eventsource.actions import Event, Fault, Start
 from ld_eventsource.errors import HTTPStatusError
 
+from ldclient.impl.aio.concurrency import AsyncTaskRunner
+from ldclient.impl.aio.transport import AsyncSSEFactory, make_client_session
 from ldclient.impl.datasource.datasource_common import (
     STREAM_ALL_PATH,
     parse_path,
     sink_or_store
 )
-from ldclient.impl.http import HTTPFactory, _http_factory
 from ldclient.impl.util import (
     http_error_message,
     is_http_error_recoverable,
@@ -32,19 +32,9 @@ from ldclient.interfaces import (
 )
 from ldclient.versioned_data_kind import FEATURES, SEGMENTS
 
-# allows for up to 5 minutes to elapse without any data sent across the stream. The heartbeats sent as comments on the
-# stream will keep this from triggering
-stream_read_timeout = 5 * 60
 
-MAX_RETRY_DELAY = 30
-BACKOFF_RESET_INTERVAL = 60
-JITTER_RATIO = 0.5
-
-
-class StreamingUpdateProcessor(Thread, UpdateProcessor):
-    def __init__(self, config, store, ready, diagnostic_accumulator):
-        Thread.__init__(self, name="ldclient.datasource.streaming")
-        self.daemon = True
+class AsyncStreamingUpdateProcessor(UpdateProcessor):
+    def __init__(self, config, store, ready, diagnostic_accumulator, sse_factory: Optional[AsyncSSEFactory] = None):
         self._uri = config.stream_base_uri + STREAM_ALL_PATH
         if config.payload_filter_key is not None:
             self._uri += '?%s' % parse.urlencode({'filter': config.payload_filter_key})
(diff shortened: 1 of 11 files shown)


ldclient/impl/datasource/async_streaming.py
```diff
@@ -0,0 +1,241 @@
...
+
+                    await self._handle_error(e)
+                except Exception as e:
+                    log.warning("Error while handling stream event; will restart stream: %s" % e)
+                    await self._sse.interrupt()
+
+                    if self._data_source_update_sink is not None:
+                        error_info = DataSourceErrorInfo(DataSourceErrorKind.UNKNOWN, 0, time.time(), str(e))
+
+                        self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, error_info)
```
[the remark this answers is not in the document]
jsonbailey: This mirrors the sync data source status sink rather than being an async-only issue: sync's `__monitor_store_update` also re-raises after setting `STORE_ERROR`, and the sync streaming `run()` has the same broad `except Exception` that then calls `update_status(INTERRUPTED, UNKNOWN)` and overwrites the more specific kind. Keeping async in parity with sync here; if the overwrite is worth correcting it should be fixed in both sync and async as a separate change. Tracked as a follow-up.

ldclient/impl/datasource/async_streaming.py
```diff
@@ -0,0 +1,241 @@
...
+            self._sse_factory = sse_factory
+            self._owned_session = None
+        else:
+            # Build a session configured from the SDK's HTTP options (CA certs,
+            # client cert, SSL verification, proxy trust) so the streaming
+            # connection isn't a plain unconfigured ClientSession. The SSE
+            # client treats the supplied session as externally owned and never
+            # closes it, so this data source closes it on teardown.
+            self._owned_session = make_client_session(config)
+            self._sse_factory = AsyncSSEFactory(config, session=self._owned_session)
```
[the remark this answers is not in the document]
jsonbailey: Leaving this as-is. The owned session is created via `make_client_session` in `__init__`, but in the wired-up path the streaming source is constructed inside the client's `async start()` (client start → data system start → `_make_update_processor`), so a running event loop is present when the session is created — aiohttp only requires a running loop, which is satisfied here, so there's no off-loop creation in practice. Deferring creation to `start()`/`_run()` would also invert the intentional eager-construction contract the ownership tests assert. If the FDv2-style per-source construction ever changes to build these off-loop, we'll revisit with lazy creation.

ldclient/impl/datasource/async_streaming.py
```diff
@@ -0,0 +1,245 @@
...
+                            self._ready.set()
+                elif isinstance(action, Fault):
+                    # If the SSE client detects the stream has closed, then it will emit a fault with no-error. We can
+                    # ignore this since we want the connection to continue.
+                    if action.error is None:
+                        continue
+
+                    if not await self._handle_error(action.error):
+                        break
+        finally:
```
joker23, line 119: Would adding a `catch` block be beneficial? In case something throws we can notify?

ldclient/impl/datasource/async_streaming.py
```diff
@@ -0,0 +1,245 @@
...
+            if self._sse:
+                await self._sse.close()
```
joker23: If this line throws then we won't get to `_close_owned_session`

ldclient/impl/datasource/async_streaming.py
```diff
@@ -0,0 +1,245 @@
...
+        elif msg.event == 'patch':
+            payload = json.loads(msg.data)
+            path = payload['path']
+            obj = payload['data']
+            log.debug("Received patch event for %s, New version: [%d]", path, obj.get("version"))
+            parsed = parse_path(path)
+            if parsed is not None:
+                await target.upsert(parsed.kind, obj)
+            else:
+                log.warning("Patch for unknown path: %s", path)
```
joker23: nit: consider sanitizing the `path` string

ldclient/impl/datasource/async_streaming.py
jsonbailey: I added a catch. When you say "notify" did you mean log or update_status? I went with logging but we can discuss this if needed.

ldclient/impl/datasource/async_streaming.py
jsonbailey: Other SDKs don't include the path in the warning so I have removed it as well to align with other sdks. I also updated the sync version which had the same issue.

ldclient/impl/datasource/async_streaming.py
joker23: yep I think logging suffices

joker23 (approved):

PR #471, opened 2026-07-27
feat: Add AsyncConfig for the async SDK client

jsonbailey:
Adds the full `AsyncConfig` class to `ldclient/async_config.py`, building on the previously-merged `AsyncBigSegmentsConfig` seed (#462). This is a stacked-PR slice of the async Python SDK epic (SDK-60), unblocked now that #463 merged `AsyncHook`/`AsyncPlugin`.

`AsyncConfig` mirrors the sync `Config` surface for the async client: base/events/stream URIs, streaming vs. polling, event buffering, private attributes, big segments, hooks, plugins, HTTP config, application metadata, and the experimental data system config. The concrete `AsyncEventProcessor` is imported only under `TYPE_CHECKING` to keep aiohttp out of the runtime import graph.

This is self-contained and reviewable on its own. It unblocks the upcoming async FDv1 polling (PR 7) and async event processor (PR 8), both of which import the full `AsyncConfig`.

The public `AsyncConfig` (and `AsyncBigSegmentsConfig`) carry an experimental `.. caution::` note — this API may change without notice and is not yet production-ready.

Tracked internally: SDK-2768

---

> [!NOTE]
> **Medium Risk**
> Large new experimental configuration surface touches networking, events, and data-source wiring for the async client, though behavior is additive and sync Config changes are limited to shared constants.
> 
> **Overview**
> Adds **`AsyncConfig`** in `ldclient/async_config.py` so **`AsyncLDClient`** can be configured with the same knobs as sync **`Config`**: endpoints, streaming/polling, events, privacy, big segments, hooks/plugins, HTTP, payload filtering, and experimental data-system settings. It wires in **`AsyncFeatureStore`** (default **`AsyncInMemoryFeatureStore`**), **`AsyncHook`** / **`AsyncPlugin`**, and factory hooks for **`AsyncUpdateProcessor`** and **`AsyncEventProcessor`**, with experimental **`.. caution::`** on the public API.
> 
> Sync **`Config`** now reuses shared **`DEFAULT_*_URI`** constants instead of inline URL literals.
> 
> **`interfaces.py`** gains experimental **`AsyncUpdateProcessor`** and **`AsyncEventProcessor`** (async **`stop`**, plus **`flush_and_wait`** on the event processor) for upcoming async polling and event delivery.
> 
[...]

PR #472, opened 2026-07-28
feat: Add async event processor

jsonbailey:
Adds the async analytics event delivery component for the async SDK client, extracted from the async SDK implementation branch (SDK-60).

Stacked on #471 (AsyncConfig) — review that first. Until #471 merges this PR shows its commits too; after merge a rebase drops them out.

## What's here

- `AsyncEventProcessor` interface + `DefaultAsyncEventProcessor` concrete implementation, mirroring the sync `EventProcessor`/`DefaultEventProcessor` split. `send_event`/`flush` are sync; `stop` is a coroutine.
- Shared sans-I/O helpers extracted into `event_processor_common.py` so the sync and async processors share buffering, output formatting, and dispatch logic.
- Tests mirroring the sync `DefaultEventProcessor` suite over an injected mock aiohttp session.

## SDK-2769

feat: Add async event processor

---

> [!NOTE]
> **Medium Risk**
> Changes analytics delivery and refactors shared sync dispatch paths; behavior is intended to match sync with broad tests, but incorrect flush/backpressure or shared-base regressions could drop or mis-deliver events.
> 
> **Overview**
> Introduces **async analytics event delivery** for `AsyncConfig` / `AsyncLDClient`: `DefaultAsyncEventProcessor` with inbox-driven `EventDispatcher`, periodic flush/context/diagnostic timers, gzip POSTs via `AsyncHTTPTransport`, and async `stop` / `flush_and_wait` semantics aligned with the sync processor.
> 
> **Shared logic** moves into `event_processor_common` as `EventDispatcherBase` (`_process_event`, indexing/dedup, debug events, `_handle_response`) plus a single `CURRENT_EVENT_SCHEMA`; the sync `EventDispatcher` now subclasses it instead of duplicating that code.
> 
> **Concurrency:** `AsyncWorkerPool` becomes **`BoundedTaskSet`** (`try_run` / `gather`-based `wait`) to cap concurrent flush POSTs; `_trigger_flush` returns whether a batch was handed off so `flush_and_wait` and shutdown can retry when all workers are busy. Tests cover shutdown flush, stop-on-shutdown-error, and a **flush_and_wait livelock** regression when workers finish in the same event-loop batch.
> 
> Adds a **large async test suite** mirroring sync `DefaultEventProcessor` behavior (headers, diagnostics, HTTP error handling, full inbox, compression).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ea75f501fb0fac6efa0a118275a3a0aeb18c651. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/impl/aio/concurrency.py
+++ b/ldclient/impl/aio/concurrency.py
@@ -1,10 +1,10 @@
 """
-Async concurrency primitives used by the async data source, event processor, and
-data system. Each wraps a piece of fiddly asyncio plumbing (timeout-aware waits,
-queue exception normalization, an interval-from-start repeating task, a bounded
-task set) that callers would otherwise inline repeatedly. The sync code uses the
-equivalent stdlib/SDK primitives (``threading.Event``/``Lock``, ``queue.Queue``,
-``RepeatingTask``, ``FixedThreadPool``) directly, so these have no sync twin.
+Async concurrency helpers used by the async data source, event processor, and
+data system. Each helper wraps one piece of asyncio setup so callers do not
+repeat it: timeout-aware waits, queue timeout/capacity exceptions, a repeating
+task, and a bounded task set. The sync code uses the standard-library and SDK
+equivalents (``threading.Event``/``Lock``, ``queue.Queue``, ``RepeatingTask``,
+``FixedThreadPool``) directly, so it has no matching helpers.
 """
 
 import asyncio
@@ -130,9 +130,9 @@ async def join_handle(handle: TaskHandle, timeout: float) -> None:
 
 
 class AsyncCallbackScheduler:
-    """Bridges sync notification paths to async callbacks: ``call`` schedules
-    a coroutine callback onto the event loop captured at construction time,
-    logging any unhandled exception. Safe to invoke from any thread."""
+    """Schedules a coroutine callback onto the event loop that was running when
+    this object was created. ``call`` runs the callback and logs any unhandled
+    exception. Safe to call from any thread."""
 
     def __init__(self):
         self._loop = asyncio.get_running_loop()
--- a/ldclient/impl/aio/concurrency.py
+++ b/ldclient/impl/aio/concurrency.py
@@ -2,7 +2,7 @@
 Async concurrency primitives used by the async data source, event processor, and
 data system. Each wraps a piece of fiddly asyncio plumbing (timeout-aware waits,
 queue exception normalization, an interval-from-start repeating task, a bounded
-task pool) that callers would otherwise inline repeatedly. The sync code uses the
+task set) that callers would otherwise inline repeatedly. The sync code uses the
 equivalent stdlib/SDK primitives (``threading.Event``/``Lock``, ``queue.Queue``,
 ``RepeatingTask``, ``FixedThreadPool``) directly, so these have no sync twin.
 """
@@ -12,7 +12,7 @@ import inspect
 import time
 from queue import Empty as QueueEmpty  # noqa: F401  (shared timeout exception)
 from queue import Full as QueueFull  # noqa: F401  (shared capacity exception)
-from typing import Any, Callable, Optional, Set
+from typing import Any, Callable, Coroutine, Optional, Set
 
 from ldclient.impl.util import log
 
@@ -250,50 +250,37 @@ class AsyncRepeatingTask:
             pass
 
 
-class AsyncWorkerPool:
-    """A fixed-size pool of concurrent tasks that rejects jobs when its limit
-    is reached. Matches the contract of
-    ``ldclient.impl.fixed_thread_pool.FixedThreadPool``."""
+class BoundedTaskSet:
+    """Runs up to ``limit`` coroutines concurrently as background tasks. When the
+    limit is reached, ``try_run`` rejects new work (returning False) rather than
+    queuing it, so callers can apply their own backpressure. ``drain`` awaits all
+    in-flight tasks; ``stop`` prevents any further work from being accepted."""
 
-    def __init__(self, size: int, name: str):
-        self._size = size
-        self._name = name
-        self._busy: Set[asyncio.Task] = set()
-        self._event = AsyncEvent()
-        self._stopped = False
+    def __init__(self, limit: int):
+        self._limit = limit
+        self._tasks: Set[asyncio.Task] = set()
+        self._accepting = True
 
-    def execute(self, jobFn: Callable) -> bool:
-        """Schedules a job for execution if the pool is not already at its
-        limit, and returns True if successful; returns False if all workers
-        are busy."""
-        if self._stopped or len(self._busy) >= self._size:
+    def try_run(self, job: Callable[[], Coroutine[Any, Any, Any]]) -> bool:
+        """Runs ``job()`` as a background task if fewer than ``limit`` tasks are
+        already running and the set is still accepting work. Returns True if the
+        task was started, or False if the set is full or has been stopped."""
+        if not self._accepting or len(self._tasks) >= self._limit:
             return False
-        task = asyncio.ensure_future(self._run_job(jobFn))
-        self._busy.add(task)
+        task = asyncio.create_task(job())
+        self._tasks.add(task)
+        task.add_done_callback(self._on_done)
         return True
 
-    async def _run_job(self, jobFn: Callable) -> None:
-        try:
-            result = jobFn()
-            if inspect.isawaitable(result):
-                await result
-        except Exception:
-            log.warning('Unhandled exception in worker thread', exc_info=True)
-        finally:
-            task = asyncio.current_task()
-            if task is not None:
-                self._busy.discard(task)
-            self._event.set()
+    def _on_done(self, task: asyncio.Task) -> None:
+        self._tasks.discard(task)
+        if not task.cancelled() and task.exception() is not None:
+            log.warning('Unhandled exception in background task', exc_info=task.exception())
 
-    async def wait(self) -> None:
-        """Waits until all currently busy workers have completed their jobs."""
-        while len(self._busy) > 0:
-            self._event.clear()
-            if len(self._busy) == 0:
-                return
-            await self._event.wait()
+    async def drain(self) -> None:
+        """Waits for all currently running tasks to complete."""
+        await asyncio.gather(*self._tasks, return_exceptions=True)
 
     def stop(self) -> None:
-        """Tells the pool to reject any further jobs; active jobs run to
-        completion."""
-        self._stopped = True
+        """Rejects any further work; tasks already running finish normally."""
+        self._accepting = False
--- a/ldclient/impl/aio/concurrency.py
+++ b/ldclient/impl/aio/concurrency.py
@@ -267,15 +267,19 @@ class BoundedTaskSet:
         task was started, or False if the set is full or has been stopped."""
         if not self._accepting or len(self._tasks) >= self._limit:
             return False
-        task = asyncio.create_task(job())
+        task = asyncio.create_task(self._run(job))
         self._tasks.add(task)
-        task.add_done_callback(self._on_done)
         return True
 
-    def _on_done(self, task: asyncio.Task) -> None:
-        self._tasks.discard(task)
-        if not task.cancelled() and task.exception() is not None:
-            log.warning('Unhandled exception in background task', exc_info=task.exception())
+    async def _run(self, job: Callable[[], Coroutine]) -> None:
+        try:
+            await job()
+        except Exception:
+            log.warning('Unhandled exception in background task', exc_info=True)
+        finally:
+            task = asyncio.current_task()
+            if task is not None:
+                self._tasks.discard(task)
 
     async def wait(self) -> None:
         """Waits for the tasks currently running to complete. This does not stop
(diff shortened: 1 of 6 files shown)


ldclient/impl/aio/concurrency.py
```diff
@@ -250,50 +250,37 @@ async def _run(self):
             pass
 
 
-class AsyncWorkerPool:
-    """A fixed-size pool of concurrent tasks that rejects jobs when its limit
-    is reached. Matches the contract of
-    ``ldclient.impl.fixed_thread_pool.FixedThreadPool``."""
+class BoundedTaskSet:
```
jsonbailey, line 253: The AsyncWorkerPool was previously created to mirror the FixedThreadPool but some of the naming and params didn't seem to align with the async code.

ldclient/impl/events/async_event_processor.py
```diff
@@ -0,0 +1,300 @@
...
+            await self._do_send(output_events)
+        except Exception:
+            log.warning('Unhandled exception in event processor. Analytics events were not processed.', exc_info=True)
+
+    async def _do_send(self, output_events):
+        # noinspection PyBroadException
+        try:
+            json_body = json.dumps(output_events, separators=(',', ':'))
+            log.debug('Sending events payload: ' + json_body)
+            payload_id = str(uuid.uuid4())
```
joker23, line 67: would this leak redacted properties?

ldclient/impl/events/async_event_processor.py
```diff
@@ -0,0 +1,300 @@
...
+                    message.param.set()
+                elif message.type == 'test_sync':
+                    await self._flush_workers.wait()
+                    if self._diagnostic_flush_workers is not None:
+                        await self._diagnostic_flush_workers.wait()
+                    message.param.set()
+                elif message.type == 'stop':
+                    await self._do_shutdown()
+                    message.param.set()
+                    return
```
joker23, line 153: If `_doShutdown` throws then it will go into the catch block, but that would not break out of the loop.

jsonbailey:
**Shutdown event-delivery parity (sync vs. async) — decision needed**

While addressing the `flush_and_wait` saturation finding, Bugbot flagged a related gap: `stop()` promises to deliver all pending events, but `_do_shutdown` never re-flushes the outbox. The pre-stop `flush()` can be dropped when the inbox is full, or left buffered when the flush workers are saturated — so those events can be abandoned on shutdown.

The proposed async fix is to have `_do_shutdown` guarantee-flush the outbox before draining workers (reusing the same retry-until-handed-off logic that now backs `flush_and_wait`).

**The catch:** this gap exists *identically* in the sync `EventProcessor` — its `stop()` / `_do_shutdown` have the same structure and can drop events the same way. Fixing only the async side would make async's shutdown more robust than sync's (a behavioral divergence), whereas today they match.

**Question:** should we fix the sync side too, and if so —
- (a) bundle the sync + async fix together (here or a combined change), keeping the two in parity, or
- (b) land the async fix here and do sync in a separate PR/follow-up?

(Leaning toward keeping this PR async-scoped and handling sync separately, but flagging for a call before applying.)

ldclient/impl/events/async_event_processor.py
jsonbailey: No — redaction happens in the shared `EventOutputFormatter.make_output_events` (in `event_processor_common`, the same code the sync event processor uses) before serialization. It strips private-attribute *values* and includes only their *names* under `_meta.redactedAttributes`. So the serialized payload — and the debug log of it — carries redacted attribute names, not values, so no redacted values leak. This is byte-identical to the sync event processor's behavior.

kinyoklion:
**This is a comment from Claude, an AI code reviewer. A human requested and reviewed this comment before it was posted.**

## Problem

The `flush_and_wait` handler can stop the event loop permanently. The dispatcher then uses approximately 100% CPU. All tasks on the application's event loop stop. Timers on that loop do not fire.

## Cause

Two behaviors cause the problem:

1. `BoundedTaskSet` releases a slot only in `_on_done`. The event loop runs `_on_done` later, through `call_soon`. A task can be complete while the set continues to count it against the limit.
2. `BoundedTaskSet.wait()` awaits `asyncio.gather(...)`. When all tasks in the set are complete, or when the set is empty, `gather` returns a future that is already complete. An `await` on a complete future does not give control back to the event loop.

The dispatcher uses this loop in the `flush_and_wait` branch of `_run_main_loop`:

```python
while not self._trigger_flush():
    await self._flush_workers.wait()
```

This sequence causes the stop:

1. The 5 POST tasks become complete in one event-loop batch.
2. The dispatcher receives the `flush_and_wait` message in the same batch.
3. `try_run` returns `False`, because the set counts the 5 complete tasks.
4. `wait()` returns immediately. It does not give control to the event loop.
5. The `_on_done` callbacks cannot run. The loop repeats without end.

With a real transport, this interleaving is a race. Socket-read completions and the inbox put only have to land in one batch. The sync SDK does not have this problem, because `FixedThreadPool` decreases its busy count synchronously in the worker thread.

## Tests

The two tests below assert the correct behavior:

- Test 1 shows the defect at the `BoundedTaskSet` level. It is small and fast. It can go into `TestBoundedTaskSet` in `ldclient/testing/test_aio.py`.
- Test 2 shows the defect end-to-end on `DefaultAsyncEventProcessor`. It can go into `ldclient/testing/impl/events/test_async_event_processor.py`.

Results on this branch (`623faa0`): the 2 tests fail in approximately 3 seconds. They do not hang the test run. A stopped event loop cannot fire its own asyncio timeouts. Because of this, test 2 uses a watchdog thread. The watchdog finds the stop after 3 seconds and clears the task set. The test then fails with a clear assertion, and CI does not hang.

```python
import asyncio
import threading

import pytest

from ldclient.async_config import AsyncConfig
from ldclient.context import Context
from ldclient.impl.aio.concurrency import AsyncEvent, BoundedTaskSet
from ldclient.impl.events.async_event_processor import (
    DefaultAsyncEventProcessor,
    EventDispatcher,
    EventProcessorMessage
)
from ldclient.impl.events.types import EventInputIdentify
from ldclient.testing.impl.events.test_async_event_processor import (
    MockAioHttp,
    MockAioResponse
)

pytestmark = pytest.mark.asyncio

context = Context.builder('userkey').name('Red').build()
timestamp = 10000


async def test_retry_pattern_makes_progress_when_a_task_finished_in_the_same_batch():
    tasks = BoundedTaskSet(1)
    finished = asyncio.Event()

    async def job():
        finished.set()

    assert tasks.try_run(job) is True

    # A raw asyncio.Event wakes this coroutine in the same loop batch in which
    # the job's task finished, *before* the task's done-callback has run.
    # This is exactly the window in which the dispatcher's flush_and_wait
    # handler can observe the set (a task completion and an inbox put landing
    # in one batch).
    await finished.wait()

    # Desired behavior: the retry pattern used by EventDispatcher._run_main_loop
    #     while not try_run(...): await wait()
    # must make progress here. Bounded to 100 iterations so the defect shows
    # up as a clean assertion failure rather than a wedged test run.
    async def noop():
        pass

    accepted = False
    for _ in range(100):
        if tasks.try_run(noop):
            accepted = True
            break
        await tasks.wait()

    assert accepted, (
        "try_run never freed capacity: wait() returned without yielding to the "
        "event loop, so the done-callback that reaps the finished task can "
        "never run; the equivalent unbounded loop in the event dispatcher "
        "spins forever"
    )
    await tasks.wait()


async def test_flush_and_wait_completes_when_in_flight_posts_finish_together():
    class GatedMockAioHttp(MockAioHttp):
        """Requests park until the gate is set, then complete without yielding."""

        def __init__(self):
            super().__init__()
            self.gate = asyncio.Event()

        def request(self, method, uri, headers=None, data=None, timeout=None, proxy=None):
            self._recorded_requests.append((headers, data))
            outer = self

            class _Ctx:
                async def __aenter__(self):
                    await outer.gate.wait()
                    return MockAioResponse(200, {})

                async def __aexit__(self, exc_type, exc_value, traceback):
                    return False

            return _Ctx()

    # Keep a handle on the dispatcher (DefaultAsyncEventProcessor discards it)
    # so the watchdog below can recover the loop if it wedges.
    dispatcher_holder = []

    def capture_dispatcher(inbox, config, http, diagnostic_accumulator):
        dispatcher = EventDispatcher(inbox, config, http, diagnostic_accumulator)
        dispatcher_holder.append(dispatcher)
        return dispatcher

    mock_http = GatedMockAioHttp()
    config = AsyncConfig(sdk_key='SDK_KEY', diagnostic_opt_out=True)
    ep = DefaultAsyncEventProcessor(config, mock_http, dispatcher_class=capture_dispatcher)
    try:
        # Saturate all 5 flush workers with parked POSTs.
        for i in range(5):
            ep.send_event(EventInputIdentify(timestamp, Context.create('user%d' % i)))
            ep.flush()
        deadline = asyncio.get_running_loop().time() + 2
        while len(mock_http.recorded_requests) < 5:
            assert asyncio.get_running_loop().time() < deadline, 'workers never saturated'
            await asyncio.sleep(0.01)

        # Buffer one more event so the awaited flush has work to hand off.
        ep.send_event(EventInputIdentify(timestamp, context))
        await asyncio.sleep(0.05)

        # The race, made deterministic: complete all in-flight POSTs and
        # enqueue the flush_and_wait message inside one loop callback, so the
        # dispatcher dequeues it in the same batch in which the workers
        # finished -- before BoundedTaskSet's done-callbacks have reaped them.
        # (With a real transport this interleaving needs no help: socket-read
        # completions and the inbox put just have to land in one batch.)
        reply = AsyncEvent()
        mock_http.gate.set()
        ep._inbox.put_nowait(EventProcessorMessage('flush_and_wait', reply))

        # While the loop is wedged, nothing scheduled on it -- including
        # asyncio timeouts -- can fire, so a plain wait_for would hang the
        # whole test run. A watchdog thread detects the wedge and forcibly
        # frees the task set so the test fails with an assertion instead.
        finished = threading.Event()
        wedged = threading.Event()

        def watchdog():
            if not finished.wait(3):
                wedged.set()
                dispatcher_holder[0]._flush_workers._tasks.clear()

        threading.Thread(target=watchdog, daemon=True).start()

        replied = await reply.wait(10)
        finished.set()

        assert replied is True
        assert not wedged.is_set(), (
            "flush_and_wait wedged the event loop: the dispatcher spun in "
            "'while not self._trigger_flush(): await self._flush_workers.wait()' "
            "without yielding, and only the watchdog thread's forcible clearing "
            "of the task set un-stuck it"
        )
    finally:
        mock_http.gate.set()
        await ep.stop()
```

## Suggested fix

Do not use the done-callbacks for the capacity count. Remove complete tasks directly in `BoundedTaskSet`:

```python
def _prune(self) -> None:
    # Do not rely on the done-callbacks (which run via call_soon) to free up
    # capacity: a task can be finished but still tracked.
    self._tasks = {t for t in self._tasks if not t.done()}

def try_run(self, job: Callable[[], Coroutine]) -> bool:
    self._prune()
    if not self._accepting or len(self._tasks) >= self._limit:
        return False
    task = asyncio.create_task(job())
    self._tasks.add(task)
    task.add_done_callback(self._on_done)
    return True

async def wait(self) -> None:
    while self._tasks:
        await asyncio.gather(*self._tasks, return_exceptions=True)
        self._prune()
```

With this fix, the 2 new tests pass. The 186 async-related tests in `ldclient/testing/test_aio.py` and `ldclient/testing/impl/events/` also pass.

**Caution**: a one-line fix, `await asyncio.sleep(0)` at the top of `wait()`, is not sufficient. The 2 new tests pass with it, but the existing test `TestBoundedTaskSet::test_saturation_returns_false` fails. The cause is the same defect in a smaller form: `wait()` can return while the set continues to count a complete task.

ldclient/impl/events/async_event_processor.py
```diff
@@ -0,0 +1,307 @@
+"""
+Implementation details of the analytics event delivery component.
+"""
+
+import asyncio
+import gzip
+import json
+import queue
+import time
```
joker23: nit: doesn't look like this package is used.

ldclient/impl/events/async_event_processor.py
```diff
@@ -0,0 +1,307 @@
...
+        hdrs['X-LaunchDarkly-Event-Schema'] = str(CURRENT_EVENT_SCHEMA)
+        hdrs['X-LaunchDarkly-Payload-ID'] = payload_id
+    can_retry = True
+    context = "posting %s" % events_description
+    data: Union[bytes, str] = gzip.compress(bytes(body, 'utf-8')) if config.enable_event_compression else body
+    while True:
+        next_action_message = "will retry" if can_retry else "some events were dropped"
+        try:
+            r = await http_client.request('POST', uri, headers=hdrs, body=data)
+            if r.status < 300:
```
joker23, line 295: nit: probably not a concern but maybe `3xx` are fine? like `304` I think is "Not modified"

joker23 (approved):
approved with nit comments

ldclient/impl/events/async_event_processor.py
jsonbailey: Good eye. We're keeping this as `< 300` to match the sync `DefaultEventProcessor`, which uses the identical check. A 3xx isn't really a case we hit here: these are unconditional event POSTs to `/bulk`, so there's no conditional request for the server to answer with a 304, and nothing we send would trigger a redirect. If we ever wanted to treat 3xx as success we'd change it in both the sync and async processors together to keep them in parity — so leaving as-is for now.

joker23 (approved):

PR #480, opened 2026-08-05
feat: Add AsyncLDClient with FDv1 data system and public API

jsonbailey:
## Summary

Adds the async LaunchDarkly client (`AsyncLDClient`) built on `asyncio`/`aiohttp`, wired to the **FDv1** data system, plus the public API surface for constructing it. This is the first extraction slice from the async SDK implementation branch (epic SDK-60).

The client uses an explicit lifecycle: construct with `AsyncLDClient(config)`, then `await start()` / `await close()` (or `async with`). Construction is **loop-free**, so the client can be built before an event loop exists and started per worker (e.g. one loop per ASGI worker).

## Lifecycle shape

- **`__init__` builds the object graph, loop-free** — data system, status providers, big-segment manager, evaluator, and flag tracker are all constructed here; no event loop is required.
- **`start()` does the loop-bound work** — creates the shared `aiohttp` session (lazily, on first use inside the loop), starts the data source, starts the big-segment status poll, and sets up the event processor.
- **`flag_tracker` and the status providers are available before `start()`**, matching the other LaunchDarkly SDKs (they no longer raise). Flag-change listeners registered before `start()` fire once data begins to flow.
- The shared HTTP session is created only when a network component first needs it, so offline/LDD mode creates none.

## What's included

- `ldclient/async_client.py` — `AsyncLDClient` with the FDv1 data system path.
- `ldclient/impl/datasystem/async_fdv1.py` — async FDv1 data system (resolves the HTTP session from a lazy provider at start).
- `ldclient/impl/datasystem/__init__.py` — adds the `AsyncDataSystem` protocol.
- `ldclient/impl/async_flag_tracker.py` — the flag tracker's callback scheduler is created lazily on first use, so the tracker can be built without a running loop.
- `ldclient/impl/async_big_segments.py` — the status poll task is created in `__init__` and started in `start()`.
- `ldclient/__init__.py`, `ldclient/client.py`, `ldclient/impl/client_common.py`, `ldclient/impl/stubs.py` — public API / shared plumbing needed by the async client (`ldclient.__getattr__` lazy-loads `AsyncLDClient` so plain `import ldclient` doesn't pull in `aiohttp`; `get_plugin_hooks` now takes `config.plugins`; async null event/update processor stubs).
- `ldclient/testing/mock_async_components.py`, `ldclient/testing/stub_util.py` — async test doubles.
- `ldclient/testing/test_async_client.py`, `ldclient/testing/impl/test_async_big_segments.py`, `ldclient/testing/test_sync_async_parity.py` — coverage for the async client, the big-segment lifecycle, and sync/async public-API parity.

## Held back for a later PR (FDv2)

This slice is **FDv1-only**. The FDv2 branch in `_make_data_system()` raises `NotImplementedError("FDv2 is not yet supported in the async client")`; the FDv2 modules (`async_fdv2`, `datasourcev2.async_*`) and their session wiring land in PR 11.

## Follow-ups

- The async **contract-test service** follows as a **stacked PR** (SDK-2868) on top of this branch.
- FDv2 support for the async client is deferred to PR 11.

## Verification

- `uv run pytest ldclient/testing --ignore=ldclient/testing/integrations -q` → 1245 passed, 2 skipped (integration suites require Redis/DynamoDB/Consul services)
- `pycodestyle`, `isort --check --atomic`, and `mypy` clean on all changed files

---

> [!NOTE]
> **Overview**
> Adds **experimental** `AsyncLDClient` for asyncio/`aiohttp` apps, mirroring the sync client’s flag evaluation, events, hooks, migrations, and status providers behind `await start()` / `await close()` (or `async with`).
> 
> **Lifecycle:** `__init__` builds the object graph without a running event loop; `start()` wires the shared HTTP session, FDv1 update processor (streaming or polling), big-segment status poll, and event processor. Failed or cancelled `start()` tears down partial startup and treats the instance as single-shot. `flag_tracker` and status providers work before `start()`; listeners registered early stay on the same data system.
> 
> **Data path:** New `AsyncDataSystem` protocol and `AsyncFDv1` implementation; FDv2 config raises `NotImplementedError`. Async null event/update processor stubs and deferred `AsyncBigSegmentStoreManager.start()` / lazy flag-tracker scheduler align with loop-free construction.
> 
> **Shared plumbing:** `get_plugin_hooks` now takes `plugins` directly (sync `LDClient` updated). `ldclient.__getattr__` lazy-loads `AsyncLDClient` so plain `import ldclient` does not require aiohttp.
> 
> Tests cover client lifecycle, hooks, `all_flags_state` degradation, big-segment start timing, and a sync/async public API parity guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e6f1bf7858b8980944cc8a8d4438c4ee467b14a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ldclient/impl/datasystem/__init__.py
```diff
@@ -143,6 +147,80 @@ def store(self) -> ReadOnlyStore:
         raise NotImplementedError
 
 
+class AsyncDataSystem(Protocol):
```
joker23, line 150: It looks like `AsyncDataSystem` class could inherit most of its structure from the `DataSystem` class and just override `store()` and `start()`. Alternatively, I think we can create an abstract base class that both types of datasystems extends from?

keelerm84 (approved):

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+
+        Safe to call multiple times — subsequent calls are no-ops.
+        """
+        async with self._lifecycle_lock:
+            if self._closed:
+                return
+            self._closed = True
+
+            if self._started:
+                try:
```
joker23: Let's make sure this doesn't open us to memory leak problem. The scenario I am thinking of is that `self._started` is either not set yet (due to waiting for start up or start up failed), then we won't be calling close on anything.

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+                # (CancelledError) also releases the components __start_up began.
+                # re-raise propagates the error/cancellation; it is not swallowed.
+                await self._cleanup_partial_start()
+                self._closed = True
+                raise
+
+            for hook in pre_start_hooks:
+                self.add_hook(hook)
+
+    async def _cleanup_partial_start(self):
```
joker23: should we also set `self._closed = true` here too? Or does it not matter since this is a pretty bad failure?

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+
+        return self._data_system.data_availability.at_least(DataAvailability.CACHED)
+
+    async def flush(self):
+        """Flushes all pending analytics events.
+
+        Normally, batches of events are delivered in the background at intervals determined by the
+        ``flush_interval`` property of :class:`ldclient.config.Config`. Calling ``flush()``
+        schedules the next event delivery to be as soon as possible; however, the delivery still
+        happens asynchronously on a worker thread, so this method will return immediately.
```
joker23: nit: I don't think `asyncio` uses worker threads? I always thought it does something similar with JS async and just runs everything on an event loop.

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+                try:
+                    await self._session.close()
+                except Exception as e:
+                    log.warning("Error closing HTTP session: %s", e)
+
+    async def _close_components(self):
+        """Releases the threads and network connections used by the SDK
+        components. The public :meth:`close` wraps this with a timeout."""
+        log.info("Closing LaunchDarkly client..")
+        await self._event_processor.stop()
```
joker23: we should probably try/catch these in case any of the `stop()` throws.

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+            pass
+        try:
+            manager = self.__big_segment_store_manager
+        except AttributeError:
+            manager = None
+        if manager is not None:
+            try:
+                await manager.stop()
+            except Exception:
+                pass
```
joker23: I think we should log something here since getting into an exception block could signal a memeory leak.

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+        log.info("Closing LaunchDarkly client..")
+        await self._event_processor.stop()
+        await self._data_system.stop()
+        await self.__big_segment_store_manager.stop()
+
+    async def __start_up(self, start_wait: float):
+        environment_metadata = get_environment_metadata(self._config, "python-server-sdk-async")
+        plugin_hooks = get_plugin_hooks(self._config.plugins, environment_metadata)
+
+        self.__hooks = self._config.hooks + plugin_hooks
```
joker23: I think you need to append to `self.__hooks` instead - I think hooks that were added via `add_hook()` before `start()` will be overwritten by this line.

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+        Customers making use of the :class:`ldclient.MigrationBuilder` should
+        not need to call this method manually.
+
+        Customers not using the builder should provide this method with the
+        tracker returned from calling :func:`migration_variation`.
+        """
+        event = tracker.build()
+
+        if isinstance(event, str):
+            log.error("error generting migration op event %s; no event will be emitted", event)
```
joker23:
```suggestion
            log.error("error generating migration op event %s; no event will be emitted", event)
```

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+        if not context.valid:
+            log.warning("Invalid context for track (%s)" % context.error)
+        else:
+            self._send_event(self._event_factory_default.new_custom_event(event_name, context, data, metric_value))
+
+    def identify(self, context: Context):
+        """Reports details about an evaluation context.
+
+        This method simply creates an analytics event containing the context properties, to
+        that LaunchDarkly will know about that context if it does not already.
```
joker23:
```suggestion
        This method simply creates an analytics event containing the context properties, so
        that LaunchDarkly will know about that context if it does not already.
```

ldclient/async_client.py
```diff
@@ -0,0 +1,747 @@
...
+        # Snapshot the hook list to ensure hooks added during evaluation don't get called for the current evaluation.
+        hooks = list(self.__hooks)  # type: List[AsyncHook]
+
+        if not hooks:
+            return await block()
+
+        series_context = EvaluationSeriesContext(key=key, context=context, default_value=default_value, method=method)
+        hook_data = await self.__execute_before_evaluation(hooks, series_context)
+        evaluation_result = await block()
+        await self.__execute_after_evaluation(hooks, series_context, hook_data, evaluation_result.evaluation_detail)
```
joker23, line 630: I think this should be in a `finally` block... in case normal code execution never reaches this line

ldclient/async_client.py
jsonbailey: Leaving these as plain sequential awaits. I checked the other server SDKs (Go, .NET, Java, JS, Ruby, sync Python) and none wrap component stops in a per-component try/catch on close — they call them sequentially and rely on the stops not throwing (only null-guards for optional components). These async stops (aiohttp `session.close()`, redis `aclose()`, the task-runner cancel/await) don't raise ordinary exceptions either; only a `CancelledError` if the outer `close()` timeout fires, which we deliberately don't swallow. So a try/catch here would be inert and would diverge from the other SDKs.

ldclient/impl/datasystem/__init__.py
jsonbailey: Good idea. The two data systems differ enough that full inheritance isn't clean (the async `stop` is a coroutine and `start` takes an async ready event), but a shared ABC for the common surface (store / status providers / `data_availability`) looks plausible. Deferring to a follow-up PR to evaluate — added to our follow-up list.

ldclient/async_client.py
jsonbailey: Leaving this sequential (no `finally`) to match the other SDKs.

`block()` is the evaluation closure, and `_evaluate_internal` always returns an `EvaluationDetail` (it has `except Exception` around both the store fetch and the evaluator call), so it can't skip this line via an ordinary exception. The only thing that propagates is `CancelledError` — and during cancellation we specifically don't want `afterEvaluation` hooks firing, which a `finally` would cause. Sync Python, Go, and JS all do `before → evaluate → after` with no `finally`.

joker23 (approved):

PR #484, opened 2026-08-11
feat: Add environment ID support for hooks.

kinyoklion:
**Requirements**

**Related issues**

Implements the `environmentId` field of `EvaluationSeriesContext` from the [hooks spec](https://github.com/launchdarkly/sdk-specs/blob/main/specs/HOOK-hooks/README.md), which the OTel tracing hook uses for `feature_flag.set.id`. Mirrors launchdarkly/dotnet-core#81.

`EvaluationSeriesContext` gains an optional `environment_id`, populated from the `X-LD-EnvID` response header sent by LaunchDarkly. Both data systems are supported, and each exposes it through a new `DataSystem.environment_id` property that the client reads when building the series context. Per the spec, it is only recorded from successful responses, so hooks never see an environment ID scraped off an error page.

- FDv2 already parsed the header into `Basis.environment_id` / `Update.environment_id` but discarded it; `FDv2` now latches the last value seen on a success path (a `_Success` basis, or an update whose state is `VALID`).
- FDv1 had no access to the header. Streaming reads it from the `Start` action of the SSE client, polling reads it from the response headers, and both record it on `DataSourceUpdateSinkImpl`.

Sketch of the FDv1 path:

```python
# streaming
if isinstance(action, Start):
    record_environment_id(self._data_source_update_sink, action.headers)

# polling
(all_data, headers) = self._get_all_data_with_headers()
record_environment_id(self._data_source_update_sink, headers)
```

Following the .NET implementation more literally, where the environment ID is stored as init metadata on the data store, would require optional extension interfaces on the public `FeatureStore`/`DataSourceUpdateSink` types. Instead, header handling stays inside the data sources and is surfaced by the data system, so externally implemented stores, sinks, feature requesters, and update processors continue to work unchanged (they simply report no environment ID).

**Additional context**

- No change is needed in `launchdarkly-eventsource`: it already exposes response headers on `Start` and `Fault` (unlike the .NET event source, which needed launchdarkly/dotnet-eventsource#104).
- The async (experimental) client does not run hooks yet, so the async data source path is untouched.
- The `environmentId` support in the OTel tracing hook lives in `python-server-sdk-otel` and is a follow-up.

Link to Devin session: https://app.devin.ai/sessions/bfe54128e2804a96bb100e6120e9a3ef
Requested by: @kinyoklion

---

> [!NOTE]
> **Overview**
> Adds optional **`environment_id`** on **`EvaluationSeriesContext`** so evaluation hooks (e.g. OTel **`feature_flag.set.id`**) can see which LaunchDarkly environment the SDK is connected to, matching the hooks spec.
> 
> The ID comes from the **`X-LD-EnvID`** header on **successful** data-source responses only—not from error responses. **FDv1** records it via streaming **`Start`** headers and polling response headers into **`DataSourceUpdateSinkImpl`**; **FDv2** latches values already present on successful **`Basis`** / **`VALID`** **`Update`** objects. **`DataSystem.environment_id`** exposes the value and **`LDClient`** passes it when building hook series context.
> 
> Custom **`FeatureRequester`** implementations without **`get_all_data_with_headers`** still work (no env ID). Contract tests add the **`hook-environment-id`** capability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ce7b6a8efffb8c24d657d16bb2114f26ac11c60. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/impl/datasource/streaming.py
+++ b/ldclient/impl/datasource/streaming.py
@@ -94,8 +94,6 @@ class StreamingUpdateProcessor(Thread, UpdateProcessor):
                         log.info("StreamingUpdateProcessor initialized ok.")
                         self._ready.set()
             elif isinstance(action, Fault):
-                record_environment_id(self._data_source_update_sink, action.headers)
-
                 # If the SSE client detects the stream has closed, then it will emit a fault with no-error. We can
                 # ignore this since we want the connection to continue.
                 if action.error is None:
--- a/ldclient/impl/datasource/streaming.py
+++ b/ldclient/impl/datasource/streaming.py
@@ -5,7 +5,7 @@ from typing import Optional
 from urllib import parse
 
 from ld_eventsource import SSEClient
-from ld_eventsource.actions import Event, Fault
+from ld_eventsource.actions import Event, Fault, Start
 from ld_eventsource.config import (
     ConnectStrategy,
     ErrorStrategy,
@@ -16,6 +16,7 @@ from ld_eventsource.errors import HTTPStatusError
 from ldclient.impl.datasource.datasource_common import (
     STREAM_ALL_PATH,
     parse_path,
+    record_environment_id,
     sink_or_store
 )
 from ldclient.impl.http import HTTPFactory, _http_factory
@@ -62,7 +63,9 @@ class StreamingUpdateProcessor(Thread, UpdateProcessor):
         self._sse = self._create_sse_client()
         self._connection_attempt_start_time = time.time()
         for action in self._sse.all:
-            if isinstance(action, Event):
+            if isinstance(action, Start):
+                record_environment_id(self._data_source_update_sink, action.headers)
+            elif isinstance(action, Event):
                 message_ok = False
                 try:
                     message_ok = self._process_message(sink_or_store(self._data_source_update_sink, self._store), action)
@@ -91,6 +94,8 @@ class StreamingUpdateProcessor(Thread, UpdateProcessor):
                         log.info("StreamingUpdateProcessor initialized ok.")
                         self._ready.set()
             elif isinstance(action, Fault):
+                record_environment_id(self._data_source_update_sink, action.headers)
+
                 # If the SSE client detects the stream has closed, then it will emit a fault with no-error. We can
                 # ignore this since we want the connection to continue.
                 if action.error is None:
--- a/ldclient/testing/impl/datasource/test_polling_processor.py
+++ b/ldclient/testing/impl/datasource/test_polling_processor.py
@@ -165,7 +165,7 @@ def test_records_environment_id_from_polling_headers():
     sink = DataSourceUpdateSinkImpl(store, Listeners(), Listeners())
     config._data_source_update_sink = sink
     setup_processor(config)
-    ready.wait()
+    assert ready.wait(2)
 
     assert sink.environment_id == 'env-abc-123'
 
@@ -177,6 +177,6 @@ def test_environment_id_is_none_when_requester_provides_no_headers():
     sink = DataSourceUpdateSinkImpl(store, Listeners(), Listeners())
     config._data_source_update_sink = sink
     setup_processor(config)
-    ready.wait()
+    assert ready.wait(2)
 
     assert sink.environment_id is None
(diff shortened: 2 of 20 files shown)


ldclient/impl/datasource/streaming.py
```diff
@@ -91,6 +94,8 @@ def run(self):
                         log.info("StreamingUpdateProcessor initialized ok.")
                         self._ready.set()
             elif isinstance(action, Fault):
+                record_environment_id(self._data_source_update_sink, action.headers)
```
kinyoklion: **[Claude] Medium: env ID is latched from error responses and exposed to hooks pre-initialization**

This `Fault`-path record means the env ID is captured from *failed* responses: per `ld_eventsource`, `Fault.headers` is populated exactly when the response was non-2xx or wrong content type. The same happens in FDv2 from failed-initializer headers (`fdv2.py:226`) and from every synchronizer update including `INTERRUPTED`/`OFF` (`fdv2.py:425`), and the value is never cleared once set.

Proven with a repro: a **401 (invalid SDK key)** response carrying `X-LD-EnvID: env-from-401` leaves the SDK permanently uninitialized (`variation()` returns defaults), yet every hook receives `environment_id == 'env-from-401'` for the process lifetime.

This diverges from both references:
- The [hooks spec](https://github.com/launchdarkly/sdk-specs/blob/main/specs/HOOK-hooks/README.md) says the environment ID is "provided by LaunchDarkly and only available **once initialization has completed**."
- The .NET implementation this PR mirrors (launchdarkly/dotnet-core#81) records it only as store init metadata written together with successfully applied data; its error handlers read only `x-ld-fd-fallback`. Under the same 401, .NET hooks get `null`.

Practical impact: any middlebox that synthesizes an error page on the SDK↔LD path (TLS-intercepting proxy, LB in front of a Relay Proxy) can permanently poison the value that downstream OTel hooks emit as `feature_flag.set.id` — even after the real connection recovers.

**Suggestion:** record only from success paths — `Start` here, the `_Success` basis in `_run_initializers` (drop the `_Fail`-headers record at `fdv2.py:226`), and gate `fdv2.py:425` on `VALID` updates — then update/remove `test_records_environment_id_from_error_response_headers` and `test_environment_id_from_initializer_error_headers`, which currently codify the error-latching. If latching from errors is intentional, it deserves an explicit callout since it's a Python-only behavioral divergence.

Rider: `FDv2._record_environment_id` (`fdv2.py:511`) only skips `None`, so a present-but-empty `X-LD-EnvID` overwrites a previously correct value; the FDv1 helper filters `''` (`datasource_common.py`). Worth applying the same `isinstance(str) and != ''` guard there for consistency.

ldclient/testing/impl/datasource/test_polling_processor.py
```diff
@@ -145,3 +145,38 @@ def verify_recoverable_http_error(http_status_code, ignore_mock):
...
+def test_records_environment_id_from_polling_headers():
+    global mock_requester
+    mock_requester = MockFeatureRequesterWithHeaders({'X-LD-EnvID': 'env-abc-123'})
+    mock_requester.all_data = {FEATURES: {}, SEGMENTS: {}}
+
+    config = Config("SDK_KEY")
+    sink = DataSourceUpdateSinkImpl(store, Listeners(), Listeners())
+    config._data_source_update_sink = sink
+    setup_processor(config)
+    ready.wait()
```
kinyoklion: **[Claude] Medium: unbounded `ready.wait()` turns a regression into a CI hang**

This wait (and the one at line 180) has no timeout. If the external-requester fallback in `PollingUpdateProcessor._get_all_data_with_headers` regresses, the resulting exception is swallowed by `_poll`'s generic handler, `ready` is never set, and the test hangs forever — the repo configures no pytest-timeout, so CI hangs until the job limit rather than reporting a failure.

Verified by mutation: replacing the fallback with an unconditional `self._requester.get_all_data_with_headers()` call makes `test_environment_id_is_none_when_requester_provides_no_headers` hang rather than fail.

**Suggestion:** `assert ready.wait(2)` in both new tests, matching the bounded waits already used elsewhere in this file (e.g. `ready.wait(0.5)` at lines 115/137).

kinyoklion (commented):
> [!NOTE]
> These comments were generated by **Claude** (multi-agent review: general / security / adversarial agents), run and posted by @rlamb.

No crash or flag-data-integrity issues found — thread safety, None-safety, backward compat, and the eventsource header contract all checked out under adversarial testing (full report available on request). Two findings worth addressing, posted inline:

1. **Environment ID is recorded from error responses and exposed to hooks before initialization** — diverges from the HOOK spec and the .NET implementation this PR mirrors (proven with a runnable repro).
2. **The two new polling tests use unbounded `ready.wait()`** — a regression in the fallback path hangs CI instead of failing (proven by mutation).

jsonbailey:
@devin rebase this PR on main and resolve the conflicts

keelerm84 (approved):

PR #485, opened 2026-08-12
feat: Add async FDv2 data sources

jsonbailey:
## Overview

Part of the async Python SDK work (epic SDK-60). This is the first of two stacked PRs
that add async FDv2 support. It adds the async FDv2 **data sources** and the shared
support they need. The follow-up PR adds the async FDv2 **data system** (coordinator)
and wires it into the async client.

This is experimental and not yet wired into a public code path on its own; the async
client only builds an `AsyncFDv2` data system in the stacked follow-up PR.

## What this PR adds

- `impl/datasourcev2/async_polling.py` — async polling data source and its builders.
- `impl/datasourcev2/async_streaming.py` — async streaming data source and its builder.
- `impl/integrations/test_datav2/async_test_data_sourcev2.py` plus a new
  `TestDataV2.async_builder` property for use with the async FDv2 data system.

## Shared refactors

To avoid duplicating parsing logic between the sync and async sources, the payload
parsing and message handling are extracted into new shared modules that both consume:

- `polling_common.py` — `polling_payload_to_changeset` / `fdv1_polling_payload_to_changeset`,
  moved out of `polling.py`. Sync `polling.py` now imports them (and re-exports them).
- `streaming_common.py` — `process_message`, moved out of `StreamingDataSource._process_message`.
  Sync `streaming.py` now calls the shared function.

Two small fixes to existing async infrastructure that the async sources depend on:

- `impl/datasource/async_status.py` — drop the read/write lock from
  `AsyncDataSourceUpdateSinkImpl`; the single-threaded event loop does not need it.
- `impl/aio/concurrency.py` — stop swallowing/re-raising `CancelledError` separately in
  `AsyncRepeatingTask`; let cancellation propagate normally.

## Testing

- `LD_SKIP_DATABASE_TESTS=1 uv run pytest ldclient/testing/impl/datasourcev2/` — 115 passed.
- `ldclient/testing/integrations/test_test_data_sourcev2.py` — 34 passed.
- `make lint` (mypy, isort, pycodestyle) — clean.

Tracked internally: SDK-2869

---

> [!NOTE]
> **Overview**
> Adds **async FDv2 data sources** for the experimental async Python SDK: `AsyncPollingDataSource` (with aiohttp requesters and FDv1 fallback builders) and `AsyncStreamingDataSource` (SSE via `AsyncSSEClient`), plus an async **TestDataV2** source exposed through `TestDataV2.async_builder`.
> 
> **Shared logic** is pulled into `polling_common.py` and `streaming_common.py` so sync `polling.py` / `streaming.py` and the new async modules share payload parsing, `map_polling_result`, FDv1 fallback signaling, and stream error classification. Sync polling now sends the selector as query param **`basis`** (was `selector`).
> 
> Small supporting changes: **`AsyncDataSourceUpdateSinkImpl`** drops its read/write lock; **`AsyncRepeatingTask`** no longer special-cases `CancelledError` inside the action loop.
> 
> Coverage includes new async polling/streaming tests and updates to polling payload parsing tests after the move to common modules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 161bb0fc74311cccec6477221351ee45f1535e0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/impl/aio/concurrency.py
+++ b/ldclient/impl/aio/concurrency.py
@@ -242,8 +242,6 @@ class AsyncRepeatingTask:
                     result = self.__action()
                     if inspect.isawaitable(result):
                         await result
-                except asyncio.CancelledError:
-                    raise
                 except Exception as e:
                     log.exception("Unexpected exception on worker task: %s" % e)
                 delay = next_time - time.time()
--- a/ldclient/impl/datasource/async_status.py
+++ b/ldclient/impl/datasource/async_status.py
@@ -3,7 +3,6 @@ from typing import Awaitable, Callable, Dict, Mapping, Optional, Set
 
 from ldclient.impl.dependency_tracker import DependencyTracker, KindAndKey
 from ldclient.impl.listeners import Listeners
-from ldclient.impl.rwlock import ReadWriteLock
 from ldclient.interfaces import (
     AsyncDataSourceUpdateSink,
     AsyncFeatureStore,
@@ -23,13 +22,11 @@ class AsyncDataSourceUpdateSinkImpl(AsyncDataSourceUpdateSink):
         self.__flag_change_listeners = flag_change_listeners
         self.__tracker = DependencyTracker()
 
-        self.__lock = ReadWriteLock()
         self.__status = DataSourceStatus(DataSourceState.INITIALIZING, time.time(), None)
 
     @property
     def status(self) -> DataSourceStatus:
-        with self.__lock.read():
-            return self.__status
+        return self.__status
 
     async def init(self, all_data: Mapping[VersionedDataKind, Mapping[str, dict]]) -> None:
         old_data: Optional[Dict[VersionedDataKind, Mapping[str, dict]]] = None
@@ -73,22 +70,21 @@ class AsyncDataSourceUpdateSinkImpl(AsyncDataSourceUpdateSink):
     def update_status(self, new_state: DataSourceState, new_error: Optional[DataSourceErrorInfo]) -> None:
         status_to_broadcast = None
 
-        with self.__lock.write():
-            old_status = self.__status
+        old_status = self.__status
 
-            if new_state == DataSourceState.INTERRUPTED and old_status.state == DataSourceState.INITIALIZING:
-                new_state = DataSourceState.INITIALIZING
+        if new_state == DataSourceState.INTERRUPTED and old_status.state == DataSourceState.INITIALIZING:
+            new_state = DataSourceState.INITIALIZING
 
-            if new_state == old_status.state and new_error is None:
-                return
+        if new_state == old_status.state and new_error is None:
+            return
 
-            self.__status = DataSourceStatus(
-                new_state,
-                self.__status.since if new_state == self.__status.state else time.time(),
-                self.__status.error if new_error is None else new_error,
-            )
+        self.__status = DataSourceStatus(
+            new_state,
+            self.__status.since if new_state == self.__status.state else time.time(),
+            self.__status.error if new_error is None else new_error,
+        )
 
-            status_to_broadcast = self.__status
+        status_to_broadcast = self.__status
 
         if status_to_broadcast is not None:
             self.__status_listeners.notify(status_to_broadcast)
--- a/ldclient/impl/datasourcev2/async_polling.py
+++ b/ldclient/impl/datasourcev2/async_polling.py
@@ -6,7 +6,6 @@ initializer, along with any required supporting classes and protocols.
 import json
 from abc import abstractmethod
 from collections import namedtuple
-from time import time
 from typing import AsyncGenerator, Mapping, Optional, Protocol, Tuple
 from urllib import parse
 
@@ -21,31 +20,26 @@ from ldclient.impl.datasource.async_feature_requester import (
     FDV1_POLLING_ENDPOINT
 )
 from ldclient.impl.datasourcev2.polling_common import (
+    PollAction,
     fdv1_polling_payload_to_changeset,
-    polling_payload_to_changeset
+    map_polling_result,
+    polling_payload_to_changeset,
+    polling_result_to_basis
 )
 from ldclient.impl.util import (
-    _LD_ENVID_HEADER,
-    _LD_FD_FALLBACK_HEADER,
     UnsuccessfulResponseException,
     _Fail,
     _headers,
     _Result,
     _Success,
-    http_error_message,
-    is_http_error_recoverable,
     log
 )
 from ldclient.interfaces import (
     AsyncInitializer,
     AsyncSynchronizer,
-    Basis,
     BasisResult,
     ChangeSet,
     ChangeSetBuilder,
-    DataSourceErrorInfo,
-    DataSourceErrorKind,
-    DataSourceState,
     Selector,
     SelectorStore,
     Update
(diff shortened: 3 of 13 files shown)


ldclient/impl/aio/concurrency.py
```diff
@@ -242,8 +242,6 @@ async def _run(self):
                     result = self.__action()
                     if inspect.isawaitable(result):
                         await result
-                except asyncio.CancelledError:
```
jsonbailey, line 245: There is no need to catch and raise here, it isn't caught by Exception so this was just extra code.

ldclient/impl/datasource/async_status.py
```diff
@@ -3,7 +3,6 @@
 
 from ldclient.impl.dependency_tracker import DependencyTracker, KindAndKey
 from ldclient.impl.listeners import Listeners
-from ldclient.impl.rwlock import ReadWriteLock
```
jsonbailey, line 6: The locks are not always needed in async python unless you are doing multiple awaits between the values you want to keep consistent which we are not doing here. It was over eager and trying to match sync too much.

ldclient/testing/impl/datasourcev2/test_async_polling.py
```diff
@@ -0,0 +1,266 @@
...
+            updates.append(update)
+            first_update_received.set()
+
+    task = asyncio.create_task(consume())
+    await asyncio.wait_for(first_update_received.wait(), timeout=2)
+    await src.stop()
+    await asyncio.wait_for(task, timeout=2)
+
+    assert len(updates) == 1
+    assert updates[0].state == DataSourceState.VALID
```
kinyoklion, line 266: **[Claude — suggested tests]**

These two tests pin the async-only loop behavior:

- `test_sync_recoverable_error_continues_polling` — makes sure one recoverable error (HTTP 500) does not stop the poll loop. The sync suite has this test; the async suite does not. Kills mutation 1.
- `test_requester_is_closed_after_stop` — makes sure the requester closes after `stop()`. The test asserts after the generator completes. Thus the test stays correct if `stop()` later becomes signal-only with a `finally` in `sync()`. Kills mutation 2.

Both tests use `asyncio.wait_for`. A regression fails in less than 2 seconds. It does not hang CI.

```suggestion
    assert updates[0].state == DataSourceState.VALID




@pytest.mark.asyncio
async def test_sync_recoverable_error_continues_polling():
    change_set = _valid_change_set()
    src = _make_source(
        [
            _Fail(error="500", exception=UnsuccessfulResponseException(500)),
            _Success(value=(change_set, {})),
        ],
        poll_interval=0.01,
    )

    gen = src.sync(_ss())
    first = await asyncio.wait_for(gen.__anext__(), timeout=2)
    second = await asyncio.wait_for(gen.__anext__(), timeout=2)
    await gen.aclose()

    assert first.state == DataSourceState.INTERRUPTED
    assert first.error is not None
    assert first.error.status_code == 500
    assert second.state == DataSourceState.VALID
    assert second.change_set is change_set


@pytest.mark.asyncio
async def test_requester_is_closed_after_stop():
    class CloseTrackingRequester(MockPollingRequester):
        def __init__(self, results):
            super().__init__(results)
            self.closed = False

        async def close(self) -> None:
            self.closed = True

    requester = CloseTrackingRequester([_Success(value=(_valid_change_set(), {}))])
    src = AsyncPollingDataSource(poll_interval=60, requester=requester)

    first_update_received = asyncio.Event()

    async def consume():
        async for _ in src.sync(_ss()):
            first_update_received.set()

    task = asyncio.create_task(consume())
    await asyncio.wait_for(first_update_received.wait(), timeout=2)
    await src.stop()
    await asyncio.wait_for(task, timeout=2)

    assert requester.closed is True
```

ldclient/testing/impl/datasourcev2/test_async_streaming.py
```diff
@@ -0,0 +1,508 @@
...
+    ) as factory_cls:
+        factory_cls.return_value.create.return_value = MockAsyncSSEClient([])
+        ss = MockSelectorStore(Selector.no_selector())
+
+        [u async for u in src.sync(ss)]
+
+    make_session.assert_not_called()
+    assert factory_cls.call_args.kwargs["session"] is supplied
+    assert supplied.closed is False
+    assert src._owned_session is None
```
kinyoklion, line 508: **[Claude — suggested tests]**

These two tests pin the async-only stream-loop behavior:

- `test_fallback_latched_on_start_carries_through_recoverable_fault` — makes sure a latched fallback directive stops the stream when a recoverable fault occurs. The sync suite has three tests for the latch; the async suite has none. Kills mutation 3.
- `test_envid_preserved_across_headerless_reconnect_start` — makes sure a reconnect `Start` without the `X-LD-EnvID` header does not clear the environment ID. Kills mutation 4.

The mock action lists are finite, so these tests cannot hang.

```suggestion
    assert src._owned_session is None




@pytest.mark.asyncio
async def test_fallback_latched_on_start_carries_through_recoverable_fault():
    """A directive latched on Start must halt the stream even on a recoverable fault."""
    src = make_streaming_data_source()

    actions = [
        Start(headers={_LD_FD_FALLBACK_HEADER: 'true'}),
        Fault(error=HTTPStatusError(503)),
        # Must never be reached — the latched directive halts the stream.
        server_intent_event(IntentCode.TRANSFER_FULL),
        put_object_event("flag-1"),
        payload_transferred_event(),
    ]

    updates = await collect_updates(src, actions)
    assert len(updates) == 1
    assert updates[0].state == DataSourceState.INTERRUPTED
    assert updates[0].error.status_code == 503
    assert updates[0].fallback_to_fdv1 is True


@pytest.mark.asyncio
async def test_envid_preserved_across_headerless_reconnect_start():
    """A reconnect Start without X-LD-EnvID must not clear the latched value."""
    src = make_streaming_data_source()

    actions = [
        Start(headers={_LD_ENVID_HEADER: 'my-env'}),
        server_intent_event(IntentCode.TRANSFER_FULL),
        put_object_event("flag-1"),
        payload_transferred_event(1),
        Start(headers={}),
        server_intent_event(IntentCode.TRANSFER_FULL),
        put_object_event("flag-2"),
        payload_transferred_event(2),
    ]

    updates = await collect_updates(src, actions)
    assert len(updates) == 2
    assert updates[0].environment_id == 'my-env'
    assert updates[1].environment_id == 'my-env'
```

kinyoklion (commented):
> [!NOTE]
> This comment is from **Claude** (a multi-agent review with general, security, and adversarial agents). @rlamb ran the review and posted the results.

## Problem: the async tests do not examine the async loop logic

Short description (ASD-STE100):

- The sync sources and the async sources contain the same loop logic in two copies.
- The shared modules (`polling_common.py`, `streaming_common.py`) have good test coverage. The two async loop copies do not.
- We did a mutation test. We changed the async loop code in four different ways. All the async tests stayed green each time.
- Thus, a defect in the async loop logic can ship, and no test will show it.

The four mutations that shipped green:

1. `async_polling.py` — change the recoverable-error branch (`WAIT_CONTINUE`) to `break`. One HTTP 500 then stops the poll loop permanently. 15/15 async polling tests pass.
2. `async_polling.py` — remove `await self._requester.close()` from `stop()`. The owned HTTP session leaks. 15/15 tests pass.
3. `async_streaming.py` — remove the `fallback_requested or` condition on the Fault path. A latched FDv1 fallback directive no longer stops the stream on a recoverable fault. 21/21 async streaming tests pass.
4. `async_streaming.py` — remove the `envid` default from the `Start` handler. A reconnect `Start` without the header clears the environment ID. 21/21 tests pass.

## Solution

The two suggestions below add four tests, one for each mutation. We made sure of this behavior:

- Each test **passes** on this branch (head `aecca5e`; full suites: 40/40 with the additions).
- Each test **fails fast** (< 1 s, no hang) when we apply its mutation.

The polling tests use `asyncio.wait_for`, so a regression cannot hang CI. Related: several existing tests in these files wait with no time limit (for example, bare `await gen.__anext__()` with `poll_interval=60`). We recommend `asyncio.wait_for` or `pytest-timeout` for those as well — a hung test stops CI and does not show a failure.

ldclient/impl/datasourcev2/async_polling.py
```diff
@@ -0,0 +1,432 @@
...
+        Fetches the data for the given selector.
+        Returns a Result containing a tuple of ChangeSet and any request headers,
+        or an error if the data could not be retrieved.
+        """
+        query_params = {}
+        if self._config.payload_filter_key is not None:
+            query_params["filter"] = self._config.payload_filter_key
+
+        if selector is not None and selector.is_defined():
+            query_params["selector"] = selector.state
```
keelerm84: I think this should be basis, not selector.

ldclient/impl/datasourcev2/async_streaming.py
```diff
@@ -0,0 +1,326 @@
...
+        # directive header of its own.
+        fallback_requested = False
+
+        try:
+            async for action in self._sse.all:
+                if isinstance(action, Fault):
+                    # If the SSE client detects the stream has closed, then it will
+                    # emit a fault with no-error. We can ignore this since we want
+                    # the connection to continue.
+                    if action.error is None:
```
keelerm84, line 169: I'm surprised we don't have to do anything in this failure case. I would think we would want to reset the changset builder if we detect any type of error.

jsonbailey: The reconnect should call the start method on the ChangeSetBuilder which resets it so doing it here isn't needed. This mirrors the sync code.

ldclient/impl/datasourcev2/async_polling.py
jsonbailey: Good eye, this was wrong in sync as well so I've fixed it in both.

keelerm84 (approved):

PR #486, opened 2026-08-12
feat: Add async FDv2 data system

jsonbailey:
## Overview

Part of the async Python SDK work (epic SDK-60). Adds the async FDv2 **data system** (coordinator) and wires it into the async client. Targets `main` (its predecessor, #485 async FDv2 data sources, has merged).

This is experimental and should not be considered production-ready.

## What this PR adds

- `impl/datasystem/async_fdv2.py` — `AsyncFDv2`, the async data system that coordinates the async initializers and synchronizers, mirrors the sync `FDv2` fallback/recovery behavior, and exposes the async data source status and flag tracking. Includes `AsyncFeatureStoreClientWrapper` for persistent-store availability polling.
- `async_client.py` wiring: `_make_data_system` builds `AsyncFDv2`; `_wire_data_source_sessions` shares the client's aiohttp session into the async data source builders so they reuse the connection pool.

## Async readiness gate (awaitable end-to-end)

Readiness and availability are now awaitable so a persistent store populated by another process (warm start / daemon) is recognized:

- `AsyncDataSystem.data_availability` and `AsyncLDClient.is_initialized()` are coroutines; the eval path awaits `data_availability()` directly.
- `AsyncFeatureStore` gains an abstract async `is_initialized()` (with `$inited` caching + monotonic latch in `AsyncCachingStoreWrapper`, surfaced via `AsyncStore.is_ready()`). The `getattr` duck-type is dropped, so a custom store that omits it fails at construction rather than silently serving defaults.
- The gate catches a persistent-store error and degrades to `DEFAULTS` rather than raising, so `variation()` never throws on a store outage. (Sync counterpart: #506.)

## Wrapper hardening

- `is_monitoring_enabled()` delegates to the store's own opt-in (matching sync), so a store that cannot report availability is not polled and left stuck unavailable.
- `close()` is idempotent, bounds the availability-poller stop with a timeout, and logs errors from the inner store close.
- `AsyncStore.close()` logs and swallows close errors instead of returning them.

## Shared refactor

- `impl/datasystem/fdv2_common.py` gains module-level `fallback_condition` / `recovery_condition`, shared by the sync and async data systems.

## Testing

- `LD_SKIP_DATABASE_TESTS=1 uv run pytest ldclient/testing/impl/datasystem/` and `ldclient/testing/test_async_client.py` — green.
- `make lint` (mypy, isort, pycodestyle) — clean.

Tracked internally: SDK-2870

---

> [!NOTE]
> **Overview**
> Adds **async Flag Delivery v2** support by introducing `AsyncFDv2`, wiring it into `AsyncLDClient` when `datasystem_config` is set (replacing the previous `NotImplementedError`), and sharing the client's aiohttp session with async polling/streaming data source builders via `_wire_data_source_sessions`.
> 
> **Readiness and availability are now async** so persistent stores populated by another process can be detected: `AsyncLDClient.is_initialized()`, `AsyncDataSystem.data_availability()`, and a new `AsyncFeatureStore.is_initialized()` (with caching/latching in `AsyncCachingStoreWrapper` and `AsyncStore.is_ready()`). Contract-test harnesses await these checks accordingly.
> 
> Sync **FDv2** is refactored to share wiring through `_FDv2Base` and module-level `fallback_condition` / `recovery_condition` in `fdv2_common.py`. `AsyncFDv2` mirrors initializer/synchronizer coordination, FDv1 fallback, persistent-store outage recovery, and adds `AsyncFeatureStoreClientWrapper` for availability polling.
> 
> `AsyncStore.close()` logs and swallows close errors instead of returning them. Large new test suites cover `AsyncFDv2`, async persistence, and readiness gating.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 111bb37125690e27e1a9c28f3e1593ae7479ac38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/async_client.py
+++ b/ldclient/async_client.py
@@ -197,6 +197,12 @@ class AsyncLDClient:
         # Start the big-segment status poll now that a loop is running.
         self.__big_segment_store_manager.start()
 
+        # FDv2 builds its data sources from builders; wire the shared session into
+        # them before starting (FDv1 pulls the session itself via its provider).
+        datasystem_config = self._config.datasystem_config
+        if datasystem_config is not None and not self._config.offline:
+            self._wire_data_source_sessions(datasystem_config)
+
         if self._config.offline:
             log.info("Started LaunchDarkly Client in offline mode")
 
@@ -221,6 +227,10 @@ class AsyncLDClient:
                 log.info("Waiting up to " + str(start_wait) + " seconds for LaunchDarkly client to initialize...")
                 await update_processor_ready.wait(start_wait)
 
+        # Warm the persistent store's initialized state so a store populated by
+        # another process is recognized before the readiness check.
+        await self._data_system.refresh_availability()
+
         if self.is_initialized() is True:
             log.info("Started LaunchDarkly Client: OK")
         else:
@@ -243,7 +253,9 @@ class AsyncLDClient:
 
             return AsyncFDv1(self._config, self._select_feature_store(), self._get_session)
 
-        raise NotImplementedError("FDv2 is not yet supported in the async client")
+        from ldclient.impl.datasystem.async_fdv2 import AsyncFDv2
+
+        return AsyncFDv2(self._config, datasystem_config)
 
     def _select_feature_store(self) -> AsyncFeatureStore:
         """Choose the async feature store for the v1 data system based on the
@@ -252,6 +264,34 @@ class AsyncLDClient:
         if feature_store is None:
             return AsyncInMemoryFeatureStore()
         return feature_store
+
+    def _wire_data_source_sessions(self, data_system_config) -> None:
+        """Provide the client's aiohttp session to any async data source
+        builders so the sources they build share the client's connection pool."""
+        from ldclient.impl.datasourcev2.async_polling import (
+            AsyncFallbackToFDv1PollingDataSourceBuilder,
+            AsyncPollingDataSourceBuilder
+        )
+        from ldclient.impl.datasourcev2.async_streaming import (
+            AsyncStreamingDataSourceBuilder
+        )
+
+        builders = list(data_system_config.initializers or []) + list(
+            data_system_config.synchronizers or []
+        )
+        if data_system_config.fdv1_fallback_synchronizer is not None:
+            builders.append(data_system_config.fdv1_fallback_synchronizer)
+
+        for builder in builders:
+            if isinstance(
+                builder,
+                (
+                    AsyncFallbackToFDv1PollingDataSourceBuilder,
+                    AsyncPollingDataSourceBuilder,
+                    AsyncStreamingDataSourceBuilder,
+                ),
+            ):
+                builder.session(self._get_session())
 
     async def __register_plugins(self, environment_metadata: EnvironmentMetadata):
         for plugin in self._config.plugins:
@@ -455,6 +495,11 @@ class AsyncLDClient:
         if self._config.offline:
             return EvaluationDetail(default, None, error_reason('CLIENT_NOT_READY')), None
 
+        # Refresh the store's initialized state only while still uninitialized;
+        # once initialized or a basis arrives, the gate reads a cached value.
+        if self._data_system.data_availability == DataAvailability.DEFAULTS:
+            await self._data_system.refresh_availability()
+
         if self._data_system.data_availability != DataAvailability.REFRESHED:
             if self._data_system.data_availability == DataAvailability.CACHED:
                 log.warning("Feature Flag evaluation attempted before client has initialized - using last known values from feature store for feature key: " + key)
@@ -525,6 +570,11 @@ class AsyncLDClient:
         if self._config.offline:
             log.warning("all_flags_state() called, but client is in offline mode. Returning empty state")
             return FeatureFlagsState(False)
+
+        # Refresh the store's initialized state only while still uninitialized;
+        # once initialized or a basis arrives, the gate reads a cached value.
+        if self._data_system.data_availability == DataAvailability.DEFAULTS:
+            await self._data_system.refresh_availability()
 
         if self._data_system.data_availability != DataAvailability.REFRESHED:
             if self._data_system.data_availability == DataAvailability.CACHED:
--- a/ldclient/async_client.py
+++ b/ldclient/async_client.py
@@ -227,11 +227,7 @@ class AsyncLDClient:
                 log.info("Waiting up to " + str(start_wait) + " seconds for LaunchDarkly client to initialize...")
                 await update_processor_ready.wait(start_wait)
 
-        # Warm the persistent store's initialized state so a store populated by
-        # another process is recognized before the readiness check.
-        await self._data_system.refresh_availability()
-
-        if self.is_initialized() is True:
+        if await self.is_initialized() is True:
             log.info("Started LaunchDarkly Client: OK")
         else:
             log.warning("Initialization timeout exceeded for LaunchDarkly Client or an error occurred. " "Feature Flags may not yet be available.")
@@ -383,18 +379,20 @@ class AsyncLDClient:
         """Returns true if the client is in offline mode."""
         return self._config.offline
 
-    def is_initialized(self) -> bool:
+    async def is_initialized(self) -> bool:
         """Returns true if the client has successfully connected to LaunchDarkly.
 
         If this returns false, it means that the client has not yet successfully connected to LaunchDarkly.
         It might still be in the process of starting up, or it might be attempting to reconnect after an
         unsuccessful attempt, or it might have received an unrecoverable error (such as an invalid SDK key)
         and given up.
+
+        This is a coroutine because determining readiness may query a persistent store.
         """
         if self.is_offline() or self._config.use_ldd:
             return True
 
-        return self._data_system.data_availability.at_least(DataAvailability.CACHED)
+        return (await self._data_system.data_availability()).at_least(DataAvailability.CACHED)
 
     async def flush(self):
         """Flushes all pending analytics events.
@@ -495,13 +493,9 @@ class AsyncLDClient:
         if self._config.offline:
             return EvaluationDetail(default, None, error_reason('CLIENT_NOT_READY')), None
 
-        # Refresh the store's initialized state only while still uninitialized;
-        # once initialized or a basis arrives, the gate reads a cached value.
-        if self._data_system.data_availability == DataAvailability.DEFAULTS:
-            await self._data_system.refresh_availability()
-
-        if self._data_system.data_availability != DataAvailability.REFRESHED:
-            if self._data_system.data_availability == DataAvailability.CACHED:
+        availability = await self._data_system.data_availability()
+        if availability != DataAvailability.REFRESHED:
+            if availability == DataAvailability.CACHED:
                 log.warning("Feature Flag evaluation attempted before client has initialized - using last known values from feature store for feature key: " + key)
             else:
                 log.warning("Feature Flag evaluation attempted before client has initialized! Feature store unavailable - returning default: " + str(default) + " for feature key: " + key)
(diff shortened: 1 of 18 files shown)


ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,624 @@
...
+                        return True
+                    continue
+
+                basis = basis_result.value
+                log.info("Initialized via %s", initializer.name)
+
+                # Apply the basis to the store
+                await self._store.apply(basis.change_set, basis.persist)
+
+                # Set ready event if and only if a selector is defined for the changeset
```
keelerm84, line 423: We are changing the spec and the logic around this. I think it's fine to leave for now, but just making sure it's on your radar.

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,624 @@
...
+                        # Check condition periodically
+                        current_status = self._data_source_status_provider.status
+                        if check_recovery and recovery_condition(current_status):
+                            return ConditionDirective.RECOVER
+                        if fallback_condition(current_status):
+                            return ConditionDirective.FALLBACK
+                    continue
+
+                log.info("Synchronizer %s update: %s", synchronizer.name, update.state)
+                if self._stop_event.is_set():
```
keelerm84, line 584: Not super likely, but a never ending queue full of "check" would prevent this from honoring the stop event.

ldclient/async_client.py
```diff
@@ -221,6 +227,10 @@ async def __start_up(self, start_wait: float):
                 log.info("Waiting up to " + str(start_wait) + " seconds for LaunchDarkly client to initialize...")
                 await update_processor_ready.wait(start_wait)
 
+        # Warm the persistent store's initialized state so a store populated by
+        # another process is recognized before the readiness check.
+        await self._data_system.refresh_availability()
```
keelerm84: This refresh runs once at startup, and `_evaluate_internal` / `all_flags_state` refresh on each call, but `is_initialized()` (line 397) reads `data_availability` with no refresh -- and cannot, because it is synchronous. That leaves the public readiness check reporting the opposite of what evaluation does, in exactly the scenario this PR sets out to fix.

Reproduced on this commit, using the PR's own `FakeAsyncCore` setup (read-only persistent store, one synchronizer configured, no basis delivered), with the store initialized by another process *after* `start()` returns:

```
1. store not yet initialized by anyone
   is_initialized() = False
2. another process just initialized the store
   is_initialized() = False        <-- expected True
   queries against store since flip: 1   (i.e. none after the flip)
3. variation() on the same client
   variation('flag-a') = 'from-store'    <-- served from the store
   is_initialized() = True               <-- only now latched, as a side effect of the eval
```

So a caller that follows the documented pattern -- check `is_initialized()`, then evaluate -- is told the client is not ready while `variation()` on the same client would have served the store's values. Worse, a caller polling `is_initialized()` to wait for readiness without ever evaluating waits forever, since nothing else drives the refresh. In sync Python the two always agree, because `data_availability` queries the store on every read.

`use_ldd` short-circuits to `True`, and pure FDv2 daemon mode (persistent store, no data sources) returns `CACHED` unconditionally, so neither of those hits this. The gap is specifically the case in the PR description: a data source is configured but has not supplied a basis yet.

More generally, this is the cost of the two-step shape -- await a refresh, then read a sync property -- which puts an invariant on every call site. If `data_availability` stays a sync property, the refresh needs to be driven by something other than the read paths; a periodic task in FDv2 would cover `is_initialized()` and evaluation both. The existing store poller only runs while the store is marked unavailable, so it is not enough on its own.

The alternative that removes the invariant altogether is to make the async availability check awaitable end to end: `AsyncDataSystem.data_availability` as `async def`, `AsyncStore.is_initialized()` as `async def`, and an awaitable initialized check on the store contract. `AsyncDataSystem` is already a separate protocol from `DataSystem`, and the async surface is unreleased and marked experimental, so the cost of moving it is near zero today and will not be later. That would also let `AsyncStore.refresh_persistent_initialized` drop the `getattr(store, "refresh_initialized", None)` duck-type check, which today silently skips the refresh for any `AsyncFeatureStore` that is not an `AsyncCachingStoreWrapper` -- a user-supplied store would report `DEFAULTS` forever with no signal. The trade is that `AsyncLDClient.is_initialized()` becomes a coroutine, which is a real API decision but an honest one for a check that does I/O.

For context on why none of this exists elsewhere: Go's `FDv2.DataAvailability()` calls `store.IsInitialized()` straight through to the persistent store, and sync Python does the same in `ldclient/impl/datasystem/fdv2.py`. The async SDK needs the extra step only because `AsyncFeatureStore.initialized` is declared a synchronous property while the real query is a coroutine.

ldclient/impl/datasystem/async_fdv1.py
```diff
@@ -124,6 +124,10 @@ async def stop(self):
     def store(self) -> AsyncReadOnlyStore:
         return self._store_view
 
+    async def refresh_availability(self) -> None:
+        """No-op; ``data_availability`` reads the store's initialized state directly."""
```
keelerm84: This comment reads as though FDv1 has no awaited state to refresh, but the state it refers to is the same stale flag FDv2 needs the refresh for. `AsyncFDv1.data_availability` reads `self._store.initialized`, which for a persistent store is `AsyncCachingStoreWrapper.initialized` -- that is `self._inited`, set only when *this* process calls `init()` or when `refresh_initialized()` runs. Since `refresh_availability` is a no-op here, `refresh_initialized()` is never called on the FDv1 path, so async FDv1 never reports `CACHED` from an externally-populated store.

Reproduced on this commit with identical config on both clients -- persistent store already populated by another process, and an `update_processor_class` that never becomes initialized (a stream that has not connected):

```
  sync  FDv1: is_initialized() = True   variation = 'from-store'
  async FDv1: is_initialized() = False  variation = 'DEFAULT'
```

So async FDv1 hands back the default where sync FDv1 serves the store's value. `use_ldd` and `offline` are unaffected -- `AsyncNullUpdateProcessor.initialized()` returns `True`, so those report `REFRESHED`. The divergence is the failover config: streaming plus a persistent store, before the stream connects.

The gap predates this PR, so fixing it here is arguably out of scope. But this PR introduces the interface member whose whole purpose is to close that class of bug, so at minimum the comment should not imply FDv1 has nothing to refresh. Either wire it up the way FDv2 does, or say plainly that async FDv1 does not yet reflect an externally-initialized store and point at a ticket.

ldclient/async_client.py
joker23: Also consider suppressing this when in `offline` mode.

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,629 @@
...
+
+        Availability polling requires the store to provide an ``is_available``
+        method so the wrapper can detect recovery.
+        """
+        return callable(getattr(self._store, "is_available", None))
+
+    async def _wrap(self, fn: Callable):
+        try:
+            return await fn()
+        except BaseException:
```
joker23, line 129: would this raise when we are in the middle of shutting down?

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,629 @@
...
+        """Stops the availability poller and closes the inner store."""
+        poller_to_stop = None
+        if not self._closed:
+            self._closed = True
+            poller_to_stop = self._poller
+            self._poller = None
+
+        if poller_to_stop is not None:
+            poller_to_stop.stop()
+            await poller_to_stop.wait_stopped()
```
joker23: we should probably put a timeout here to prevent hanging on this coroutine.

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,629 @@
...
+
+        close = getattr(self._store, "close", None)
+        if callable(close):
+            result = close()
```
joker23: probably not a big deal, but could we make this idepotent? In case `close()` is called multiple times... not sure if that would cause issues with the store.

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,629 @@
...
+                try:
+                    await self._active_synchronizer.stop()
+                except Exception as e:
+                    log.error("Error stopping active data source: %s", e)
+
+        # Wait for the coordinator's background work to complete
+        await self._runner.stop_all(timeout=5.0)
+
+        # Close the store
+        await self._store.close()
```
joker23, line 331: nit: log the exception raised by `close()` in case that is the root cause of some leak.

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -0,0 +1,629 @@
...
+    async def _consume_synchronizer_results(
+        self,
+        synchronizer: AsyncSynchronizer,
+        set_on_ready: AsyncEvent,
+        check_recovery: bool,
+    ) -> ConditionDirective:
+        """
+        Consume results from a synchronizer until a condition is met or it fails.
+
+        :return: the ConditionDirective describing how to proceed
```
joker23, line 541: nit: return type is incorrect.

ldclient/impl/datasystem/async_fdv2.py
jsonbailey: It can re-raise, but by design and without harm during shutdown. `_wrap` propagates the store error to whoever issued the operation — the same path as any store failure — and an in-flight evaluation already degrades to the default on a store error rather than crashing. It won't cause churn either: `_update_availability` early-returns once `_closed` is set, so a late failure can't spin up a recovery poller, and `close()` is now idempotent and time-bounded.

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -110,12 +112,16 @@ def disable_cache(self) -> None:
...
-        """Returns whether the inner store supports availability checks.
+        """Returns whether the inner store opts in to availability monitoring.
 
-        Availability polling requires the store to provide an ``is_available``
-        method so the wrapper can detect recovery.
+        Delegates to the store's own opt-in. A store that does not report
+        availability is not polled, so it is never marked unavailable with no
+        path back to recovery.
         """
-        return callable(getattr(self._store, "is_available", None))
```
joker23, line 118: Are we checking for the store `is_available` elsewhere now?

ldclient/impl/datasystem/async_fdv2.py
```diff
@@ -161,22 +167,31 @@ async def _check_availability(self) -> None:
...
-        poller_to_stop = None
-        if not self._closed:
-            self._closed = True
-            poller_to_stop = self._poller
-            self._poller = None
+        """Stops the availability poller and closes the inner store.
 
+        Does nothing on a later call, so closing more than once is safe.
+        """
+        if self._closed:
```
joker23, line 174: I think this could get us into another situation where we never close since the boolean is set before any of the actual closing happens... this might be fine if we think that calling the `stop`/`close` functions are our best effort.

ldclient/async_client.py
```diff
@@ -383,18 +379,20 @@ def is_offline(self) -> bool:
         """Returns true if the client is in offline mode."""
         return self._config.offline
 
-    def is_initialized(self) -> bool:
+    async def is_initialized(self) -> bool:
```
joker23, line 382: not sure if this is a problem, but does python do truthy so when `is_initialized()` returns a future then `if client.is_initialized()` will always test true?

ldclient/impl/datasystem/async_fdv2.py
jsonbailey: Correct, and this is best-effort by design. We set the flag first so a second `close()` call does nothing. Both teardown steps have their own guards, so neither one raises. Only task cancellation can stop the close early, which means an abnormal shutdown. If you want a retry after a failed close, we would need a lock so callers wait for the first close — more than a shutdown path needs, but I can add it if you prefer.

ldclient/async_client.py
jsonbailey: `is_initialized()` is now a coroutine, so `if client.is_initialized()` is always true unless you `await` it. This applies to every async method here, such as `variation` and `all_flags_state`. The fix is to `await` the call. The docstring now states that it is a coroutine, and all internal calls await it.

ldclient/impl/datasystem/async_fdv2.py
jsonbailey: Yes. The recovery poller still calls `is_available`, in `_check_availability`. This method now decides only *whether* to monitor. It uses the store's own opt-in (`is_monitoring_enabled`), which is true when the core provides `is_available`.

ldclient/impl/datasystem/async_fdv2.py
jsonbailey: We now log it in the close method and don't return the exception as it was never used in async or sycn fdv2.

keelerm84 (approved):

joker23 (approved):

PR #488, opened 2026-08-13
feat: Add async persistent feature store foundation and Redis adapter

jsonbailey:
BEGIN_COMMIT_OVERRIDE
feat: Add async persistent feature store foundation and Redis adapter
fix: Cap the Redis feature store upsert retry loop at 10 attempts
END_COMMIT_OVERRRIDE

This PR adds the foundation for async persistent feature stores to the async server SDK, with Redis as the first adapter. Until now the async SDK had only an in-memory feature store and an async Redis Big Segment store; there was no async persistent feature store and no async caching layer. This provides that foundation, designed to be adapter-agnostic so DynamoDB and Consul adapters can be stacked on top later.

## What this PR adds

- **`AsyncFeatureStoreCore` protocol** (`ldclient/interfaces.py`) — the async equivalent of `FeatureStoreCore`: the simplified, database-specific surface (`get_internal`, `get_all_internal`, `init_internal`, `upsert_internal`, `initialized_internal`) that an adapter implements. Optional `is_available`/`close` are detected by the wrapper rather than declared, matching the sync core.
- **`_CachingStoreWrapperBase`** (`ldclient/feature_store_helpers.py`) — a shared, sans-I/O cache base holding the cache setup and the cache bookkeeping over `self._cache`. Both the sync `CachingStoreWrapper` and the new async wrapper subclass it, so the caching logic lives once. Each subclass owns its precisely-typed `_core` and its I/O-shaped methods.
- **`AsyncCachingStoreWrapper`** (`ldclient/async_feature_store_helpers.py`) — implements `AsyncFeatureStore` by wrapping an `AsyncFeatureStoreCore` plus an async-safe cache. Mirrors the sync cache semantics (disabled / TTL / infinite; init/get/all/upsert/delete; decoded model objects), awaiting the core. `upsert`/`delete` return a `bool` per the async store contract; `initialized` reflects whether `init` has completed in this process (the interface property is synchronous, so it cannot run a coroutine store query).
- **Async Redis adapter** (`ldclient/impl/integrations/redis/async_redis_feature_store.py`) — `_AsyncRedisFeatureStoreCore`, mirroring the sync `_RedisFeatureStoreCore` using `redis.asyncio` (the same client the async Big Segment store uses), including the WATCH/MULTI optimistic-lock upsert with retry.
- **Integration entry point** — `Redis.async_feature_store(...)` (`ldclient/integrations/__init__.py`), following the existing `Redis.async_big_segment_store` naming.
- **Same Redis key layout** as the sync store (`{prefix}:features` hash + `{prefix}:$inited` marker), so an async and a synchronous SDK can share one Redis instance.

## Testing

- New non-DB unit tests for the caching-wrapper logic (`ldclient/testing/test_async_feature_store_helpers.py`) using an in-memory mock core — these run without Redis.
- New DB-gated integration tests for the async Redis feature store (`ldclient/testing/integrations/test_async_redis_feature_store.py`) — gated by `LD_SKIP_DATABASE_TESTS`, mirroring the existing async Redis Big Segment tests, plus the two upsert race-condition tests adapted to async.
- Full unit suite green: `LD_SKIP_DATABASE_TESTS=1 python -m pytest ldclient/testing` → **1363 passed, 279 skipped**. The Redis code paths were also exercised locally against `fakeredis` (init/get/all/upsert incl. version-rejection/delete/is_available and both WATCH-race scenarios).
- mypy, isort, and pycodestyle clean on the touched files.

## Also caps the sync Redis store
The sync Redis feature store (`redis_feature_store.py`) had the same unbounded WATCH-retry loop. This PR caps it too (10 attempts, raise on exhaustion) — note this is a behavior change to already-released code: an unbounded spin under sustained contention becomes a surfaced store error.

Tracked internally: SDK-2905

---

> [!NOTE]
> **Overview**
> Adds an **experimental** async persistent feature store stack for `AsyncLDClient`, mirroring the sync `FeatureStoreCore` + `CachingStoreWrapper` pattern.
> 
> **Shared caching:** `CachingStoreWrapper` now subclasses `_CachingStoreWrapperBase`, which centralizes cache setup, `disable_cache`, and the get/put/init/upsert cache helpers; sync methods delegate to those helpers instead of duplicating logic.
> 
> **Async surface:** New `AsyncFeatureStoreCore` protocol and `AsyncCachingStoreWrapper` implement `AsyncFeatureStore` with the same caching semantics as sync (including `upsert`/`delete` returning whether the write applied). `initialized` only reflects whether `init` completed in-process, since the property cannot await a store query.
> 
> **Redis:** `_AsyncRedisFeatureStoreCore` uses `redis.asyncio` with the same key layout as the sync Redis store (WATCH/MULTI upsert with bounded retries). `Redis.async_feature_store(...)` wires core + wrapper for use in client config.
> 
> **Tests:** Mock-core unit tests for the async wrapper and DB-gated integration tests for async Redis (prefix isolation, upsert races, availability).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f4d0f747ba3a6a95973f943f108eed782872437. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/impl/integrations/redis/async_redis_feature_store.py
+++ b/ldclient/impl/integrations/redis/async_redis_feature_store.py
@@ -1,0 +1,118 @@
+import json
+from typing import Any, Callable, Dict, Mapping, Optional
+
+from ldclient.feature_store_helpers import CachingStoreWrapper
+from ldclient.impl.util import log, redact_password
+from ldclient.interfaces import AsyncFeatureStoreCore, DiagnosticDescription
+from ldclient.versioned_data_kind import VersionedDataKind
+
+have_async_redis = False
+try:
+    import redis.asyncio as redis_client
+    from redis.exceptions import WatchError
+
+    have_async_redis = True
+except ImportError:
+    pass
+
+
+class _AsyncRedisFeatureStoreCore(DiagnosticDescription, AsyncFeatureStoreCore):
+    """Async Redis implementation of :class:`ldclient.interfaces.AsyncFeatureStoreCore`.
+
+    It stores data in the same Redis key layout as the synchronous Redis feature store, so an async
+    and a synchronous SDK can share one Redis instance.
+    """
+
+    def __init__(self, url: str, prefix: Optional[str], redis_opts: Dict[str, Any]):
+        if not have_async_redis:
+            raise NotImplementedError("Cannot use async Redis feature store because redis package is not installed")
+        self._prefix = prefix or 'launchdarkly'
+        self._init_key = "{0}:{1}".format(self._prefix, CachingStoreWrapper.__INITED_CACHE_KEY__)
+        self._client = redis_client.from_url(url, **redis_opts)
+        self.test_update_hook: Optional[Callable[[str, str], None]] = None  # exposed for testing
+        log.info("Started AsyncRedisFeatureStore connected to URL: " + redact_password(url) + " using prefix: " + self._prefix)
+
+    async def is_available(self) -> bool:
+        try:
+            await self.initialized_internal()
+            return True
+        except BaseException:
+            return False
+
+    def _items_key(self, kind: VersionedDataKind) -> str:
+        return "{0}:{1}".format(self._prefix, kind.namespace)
+
+    async def init_internal(self, all_data: Mapping[VersionedDataKind, Mapping[str, dict]]) -> None:
+        all_count = 0
+        async with self._client.pipeline() as pipe:
+            for kind, items in all_data.items():
+                base_key = self._items_key(kind)
+                pipe.delete(base_key)
+                for key, item in items.items():
+                    pipe.hset(base_key, key, json.dumps(item))
+                all_count = all_count + len(items)
+            pipe.set(self._init_key, self._init_key)
+            await pipe.execute()
+        log.info("Initialized AsyncRedisFeatureStore with %d items", all_count)
+
+    async def get_all_internal(self, kind: VersionedDataKind) -> Mapping[str, dict]:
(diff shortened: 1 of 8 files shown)


ldclient/impl/integrations/redis/async_redis_feature_store.py
```diff
@@ -0,0 +1,118 @@
...
+            log.debug("AsyncRedisFeatureStore: key %s not found in '%s'. Returning None.", key, kind.namespace)
+            return None
+        return json.loads(item_json.decode('utf-8'))
+
+    async def upsert_internal(self, kind: VersionedDataKind, item: dict) -> dict:
+        base_key = self._items_key(kind)
+        key = item['key']
+        item_json = json.dumps(item)
+
+        while True:
```
keelerm84: Do we have a max number of retries for this? Just did this in one of the go packages and they both have 10 as the retry limit.

keelerm84 (approved):

keelerm84 (approved):

ldclient/impl/integrations/redis/async_redis_feature_store.py
jsonbailey: Added — capped at 10 attempts (matching the Go Redis stores) and it raises on exhaustion, so `upsert_internal` gives up instead of spinning forever and the SDK treats the store as unavailable and refreshes on recovery. The sync Redis store had the same unbounded loop, so I applied the same cap there in this PR too.

PR #490, opened 2026-08-14
feat: Add async DynamoDB persistent feature store

jsonbailey:
## Overview

Adds an async DynamoDB persistent feature store: `AsyncDynamoDBFeatureStore`,
constructed via `DynamoDB.async_feature_store(...)`. It builds directly on the async
persistence foundation merged in #488 (`AsyncFeatureStoreCore` + `AsyncCachingStoreWrapper`),
providing the DynamoDB-backed core.

The store uses the same table layout and key schema as the synchronous DynamoDB feature
store, so an async SDK and a synchronous SDK can share a single DynamoDB table.

## New optional dependency

Adds an `async-dynamodb` extra that pulls in `aioboto3`. `boto3` is sync-only, so async
access needs `aioboto3` (which runs on `aiobotocore`). The existing sync `dynamodb` extra
(boto3) is unchanged.

## Client lifecycle

`aioboto3` clients are async context managers and cannot be created in `__init__` the way
the sync boto3 client is. The client is therefore created lazily on first use, bound to the
running event loop, kept for the lifetime of the store, and released in `close()`.

## Testing

Unit tests that do not need a backend (the caching-wrapper wiring and the missing-`aioboto3`
guard) run in the normal suite. The integration tests require a live DynamoDB Local instance,
so they run in CI rather than locally.

## Notes

- SDK-2906
- Draft pending review.

feat: Add async DynamoDB persistent feature store

---

> [!NOTE]
> **Overview**
> Adds **`DynamoDB.async_feature_store(...)`** for `AsyncLDClient`, wrapping a new **`_AsyncDynamoDBFeatureStoreCore`** with **`AsyncCachingStoreWrapper`**. The core mirrors the sync DynamoDB layout (JSON `item`, versioned upserts, non-atomic `init` with post-hoc deletes) but uses **aioboto3** with lazy client creation under an **`asyncio.Lock`**, **`AsyncExitStack`** lifecycle, and **`close()`** that prevents new clients after shutdown.
> 
> Documents the feature as **experimental**, adds the **`async-dynamodb`** optional extra (`aioboto3>=11.0`), and includes integration tests against DynamoDB Local plus unit tests for wiring, missing **aioboto3**, and post-close client access.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4dc3cb616c14b11352a897885d185b9df480e4cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/impl/integrations/dynamodb/async_dynamodb_feature_store.py
+++ b/ldclient/impl/integrations/dynamodb/async_dynamodb_feature_store.py
@@ -48,12 +48,6 @@ except ImportError:
 
 
 class _AsyncDynamoDBFeatureStoreCore(DiagnosticDescription, AsyncFeatureStoreCore):
-    """Async DynamoDB implementation of :class:`ldclient.interfaces.AsyncFeatureStoreCore`.
-
-    It stores data in the same DynamoDB table layout and key schema as the synchronous DynamoDB
-    feature store, so an async and a synchronous SDK can share one table.
-    """
-
     PARTITION_KEY = 'namespace'
     SORT_KEY = 'key'
     VERSION_ATTRIBUTE = 'version'
--- a/ldclient/impl/integrations/dynamodb/async_dynamodb_feature_store.py
+++ b/ldclient/impl/integrations/dynamodb/async_dynamodb_feature_store.py
@@ -1,0 +1,220 @@
+import asyncio
+import json
+from contextlib import AsyncExitStack
+from typing import Any, Mapping, Optional, cast
+
+from ldclient.impl.util import log
+from ldclient.interfaces import AsyncFeatureStoreCore, DiagnosticDescription
+from ldclient.versioned_data_kind import VersionedDataKind
+
+have_aioboto3 = False
+try:
+    import aioboto3
+
+    have_aioboto3 = True
+except ImportError:
+    pass
+
+
+#
+# Internal implementation of the async DynamoDB feature store.
+#
+# Implementation notes:
+#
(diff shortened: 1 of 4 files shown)


keelerm84 (approved):

ldclient/impl/integrations/dynamodb/async_dynamodb_feature_store.py
```diff
@@ -61,13 +61,17 @@ def __init__(self, table_name: str, prefix: Optional[str], dynamodb_opts: Mappin
...
+        # Guards lazy client creation and close so they cannot interleave: a
+        # close must not race a client that is still being created.
         self._client_lock = asyncio.Lock()
 
     async def _get_client(self) -> Any:
         if self._client is not None:
             return self._client
         async with self._client_lock:
+            if self._closed:
+                raise RuntimeError("DynamoDB feature store is closed")
```
keelerm84, line 74: I'm not a fan of raising an exception, but this shouldn't ever happen in practice right?

keelerm84 (approved):

ldclient/impl/integrations/dynamodb/async_dynamodb_feature_store.py
jsonbailey: This should never happen and if it did, the exception would be treated as store unavailable upstream. I can switch it if needed, but it should be safe.

PR #491, opened 2026-08-14
feat: Add async Consul persistent feature store

jsonbailey:
## Overview

Adds an async Consul persistent feature store, built on the async persistence foundation introduced in #488 (`AsyncFeatureStoreCore` + `AsyncCachingStoreWrapper`).

`Consul.async_feature_store(...)` returns an `AsyncCachingStoreWrapper` around the new `_AsyncConsulFeatureStoreCore`, giving async SDK users a Consul-backed implementation of `AsyncFeatureStore`. The store uses the same Consul KV layout as the synchronous Consul feature store, so an async and a synchronous SDK can share one Consul instance.

## Implementation notes

- The underlying `consul.aio.Consul` client builds its aiohttp session at construction time, which needs a running event loop. The core creates the client **lazily**, bound to the running event loop on the first store operation, so the factory method can still be called synchronously while building configuration. The client is released in `close()`.

## New optional dependency

- Adds the `async-consul` extra, which pulls in `py-consul[asyncio]` — the maintained Criteo fork of `python-consul` that provides an asyncio client over aiohttp.
- The **synchronous** Consul store still depends on the older `python-consul`, so the SDK temporarily depends on two Consul packages. Consolidating onto a single package is tracked separately in SDK-2913.

## Testing

- Wrapper/caching logic is covered without a live backend by `test_async_feature_store_helpers`.
- The Consul integration tests require a live Consul agent, so they run in CI.

## References

- SDK-2907
- Builds on #488

---

Draft pending review.

### DIFF
--- a/ldclient/impl/integrations/consul/async_consul_feature_store.py
+++ b/ldclient/impl/integrations/consul/async_consul_feature_store.py
@@ -1,0 +1,162 @@
+import json
+from typing import Any, Dict, Mapping, Optional
+
+from ldclient import log
+from ldclient.interfaces import AsyncFeatureStoreCore, DiagnosticDescription
+from ldclient.versioned_data_kind import VersionedDataKind
+
+have_async_consul = False
+try:
+    import consul.aio
+
+    have_async_consul = True
+except ImportError:
+    pass
+
+#
+# Internal implementation of the async Consul feature store.
+#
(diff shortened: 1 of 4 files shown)


jsonbailey:
We are holding off on async consul support for the moment. If there is a customer need we will reconsider.

PR #494, opened 2026-08-15
fix: Bound the shutdown wait so close() cannot block forever

### LINKED ISSUE #493: LDClient.close() blocks forever when DNS resolution hangs
## Describe the bug

`LDClient.close()` can block forever. There is no timeout anywhere on its shutdown path, so if delivery of the final analytics event payload stalls, `close()` never returns and the calling thread is stuck permanently.

The trigger we hit in production is **hanging DNS resolution**. This is not covered by `HTTPConfig(connect_timeout=..., read_timeout=...)`: urllib3 calls `socket.getaddrinfo()` in `util/connection.py` *before* it applies `sock.settimeout()`, so name resolution is outside both timeouts. When a resolver blackholes requests rather than returning NXDOMAIN, a flush worker sits in `getaddrinfo()` indefinitely and the whole shutdown path queues up behind it.

This bit us hard because `close()` is typically called from an `atexit` / interpreter-shutdown hook. A process that can never finish `close()` can never exit. We had worker pods sitting alive for **hours to days** after finishing their work, each holding a concurrency slot that was never released.

### The unbounded path

Every wait in this chain is untimed:

1. `LDClient.close()` → `self._event_processor.stop()` — `ldclient/client.py:356`
2. `DefaultEventProcessor.stop()` → `self._post_message_and_wait('stop')` — `ldclient/impl/events/event_processor.py:212`
3. `_post_message_and_wait()` → `reply.wait()` with no timeout — `event_processor.py:233-236`. Note this method *also* does a blocking `self._inbox.put(...)` on a bounded queue, which is a second unbounded wait.
4. the reply is only set after `_do_shutdown()` returns — `event_processor.py:150-151`
5. `_do_shutdown()` → `self._flush_workers.wait()` — `event_processor.py:163-165`
6. `FixedThreadPool.wait()` → `self._event.wait()` with no timeout — `ldclient/impl/fixed_thread_pool.py:41-47`
7. which only returns once the in-flight `EventPayloadSendTask` finishes — and that is the task stuck in `getaddrinfo()`.

## To reproduce

This uses the real SDK and the real urllib3 stack. The only thing faked is the resolver — `socket.getaddrinfo` is made to hang for the events host, which is exactly what a blackholed resolver looks like to the process. Note `connect_timeout` and `read_timeout` are both set to 1 second and do not help.

```python
import socket, sys, threading, time

HANG_SECONDS = 600
_real_getaddrinfo = socket.getaddrinfo

def hanging_getaddrinfo(host, port, *args, **kwargs):
    if "launchdarkly" in str(host):
        print(f"    [resolver] getaddrinfo({host!r}) -> hanging (simulated DNS blackhole)", flush=True)
        time.sleep(HANG_SECONDS)
    return _real_getaddrinfo(host, port, *args, **kwargs)

socket.getaddrinfo = hanging_getaddrinfo

from ldclient.config import Config, HTTPConfig
from ldclient.context import Context
from ldclient.impl.events.event_processor import DefaultEventProcessor
from ldclient.impl.events.types import EventInputIdentify
from ldclient.version import VERSION

config = Config(
    "fake-sdk-key",
    use_ldd=True,
    diagnostic_opt_out=True,
    http=HTTPConfig(connect_timeout=1, read_timeout=1),
)

print(f"ldclient version: {VERSION}", flush=True)

ep = DefaultEventProcessor(config)
ep.send_event(EventInputIdentify(int(time.time() * 1000), Context.create("user-key")))
ep.flush()
time.sleep(1.0)  # let the flush worker pick up the payload and get stuck

print("calling stop() -- this is what LDClient.close() does first", flush=True)
finished = threading.Event()
started = time.time()
threading.Thread(target=lambda: (ep.stop(), finished.set()), daemon=True).start()

if finished.wait(20.0):
    print(f"RESULT: stop() returned after {time.time() - started:.1f}s -- BOUNDED", flush=True)
    sys.exit(0)
print("RESULT: stop() has not returned after 20s -- UNBOUNDED (bug reproduced)", flush=True)
sys.exit(1)
```

### Output on 9.16.1

```
ldclient version: 9.16.1
    [resolver] getaddrinfo('events.launchdarkly.com') -> hanging (simulated DNS blackhole)
calling stop() -- this is what LDClient.close() does first
RESULT: stop() has not returned after 20s and is waiting on a 600s resolver -- UNBOUNDED (bug reproduced)
```

The thread blocked inside `close()`:

```
  File "repro.py", line 62, in <lambda>
    threading.Thread(target=lambda: (ep.stop(), finished.set()), daemon=True).start()
  File "ldclient/impl/events/event_processor.py", line 212, in stop
    self._post_message_and_wait('stop')
  File "ldclient/impl/events/event_processor.py", line 236, in _post_message_and_wait
    reply.wait()
  File "threading.py", line 669, in wait
    signaled = self._cond.wait(timeout)
```

and the flush worker it is waiting on:

```
  File "ldclient/impl/fixed_thread_pool.py", line 63, in _run_worker
    item()
  File "ldclient/impl/events/event_processor.py", line 54, in run
    resp = self._do_send(output_events)
  File "ldclient/impl/events/event_processor.py", line 261, in _post_events_with_retry
    r = http_client.request('POST', uri, headers=hdrs, body=data, timeout=urllib3.Timeout(...), retries=0)
  ...
  File "urllib3/util/connection.py", line 60, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
```

## Expected behavior

`close()` should always return within a bounded time. Analytics events are already best-effort — they are dropped when the inbox is full, when the outbox overflows, and when all flush workers are busy — so dropping a final payload that cannot be delivered is consistent with existing behavior, and far better than never returning.

## Logs

Nothing is logged; the SDK simply never returns from `close()`.

## SDK version

9.16.1 (also present in 9.15.x and 9.16.0; the relevant code is unchanged on `main` as of 5da1515)

## Language version, developer tools

Python 3.12 and 3.14, urllib3 2.7.0

## OS/platform

Linux (reproduced on 6.18 x86_64). Originally hit on Kubernetes, where a degraded cluster resolver made this routine.

## Additional context

`DefaultAsyncEventProcessor.stop()` has the same unbounded shape — `await self._post_message_and_wait('stop')` with no timeout (`ldclient/impl/events/async_event_processor.py:245`), while `flush_and_wait()` directly above it already bounds its wait with `asyncio.wait_for`. I have not reproduced a hang there, so I left it alone, but it looks worth a look.

I have opened a PR fixing the sync path: it adds a `shutdown_timeout` config option (defaulting to 5 seconds) and threads it through the shutdown waits.

aviadr1:
Fixes #493.

## The bug

`LDClient.close()` has no timeout anywhere on its shutdown path. If delivery of the final analytics event payload stalls, `close()` never returns.

The trigger is **hanging DNS resolution**, which `HTTPConfig`'s `connect_timeout` and `read_timeout` do not cover: urllib3 calls `socket.getaddrinfo()` in `util/connection.py` *before* it applies `sock.settimeout()`, so name resolution sits outside both. When a resolver blackholes requests instead of returning NXDOMAIN, a flush worker blocks in `getaddrinfo()` indefinitely and the entire shutdown path queues up behind it.

Because `close()` is usually called from an `atexit` / interpreter-shutdown hook, a process in this state can never exit. We hit this on Kubernetes with a degraded cluster resolver and had worker pods alive for hours to days after finishing their work, each holding a concurrency slot that was never released.

Every wait in the chain was untimed:

```
DefaultEventProcessor.stop()
  -> _post_message_and_wait('stop')   # blocking put on a bounded queue, then reply.wait()
  -> EventDispatcher._do_shutdown()   # runs before the reply is set
  -> FixedThreadPool.wait()           # Event.wait() with no timeout
  -> the EventPayloadSendTask stuck in getaddrinfo()
```

Note there are two unbounded waits in `_post_message_and_wait` alone — the blocking `Queue.put` as well as `reply.wait()` — so bounding only the reply would still leave a way to hang.

## The fix

Add a `shutdown_timeout` config option, defaulting to 5 seconds, and thread it through those waits:

- `FixedThreadPool.wait()` takes an optional timeout and returns whether it actually drained.
- `_post_message_and_wait()` takes an optional timeout, bounding both the inbox put and the reply wait, and returns whether the message was handled.
- `_do_shutdown()` shares a single deadline across its flush-worker and diagnostic-worker waits, so they cannot add up to more than the budget.

When the budget is exhausted the SDK logs a warning and returns. Worker threads are daemons, so any left stuck do not keep the process alive.

## On the default

I made the default bounded (5 seconds, matching `flush_interval`) rather than preserving the current unbounded wait, because an unbounded default is what makes this a hang rather than a delay, and the failure mode is silent and unrecoverable.

Dropping an undeliverable final payload is consistent with how analytics events are already treated — they are dropped when the inbox is full, when the outbox overflows, and when all flush workers are busy. Anyone who wants the old behavior can set `shutdown_timeout=None`.

Happy to change the default or the option name if you'd prefer something else.

## Verification

Reproduction using the real SDK and real urllib3 stack, with only `socket.getaddrinfo` made to hang (full script in #493):

```
# before
ldclient version: 9.16.1
    [resolver] getaddrinfo('events.launchdarkly.com') -> hanging (simulated DNS blackhole)
calling stop() -- this is what LDClient.close() does first
RESULT: stop() has not returned after 20s and is waiting on a 600s resolver -- UNBOUNDED (bug reproduced)

# after
ldclient version: 9.16.1
    [resolver] getaddrinfo('events.launchdarkly.com') -> hanging (simulated DNS blackhole)
calling stop() -- this is what LDClient.close() does first
Timed out waiting for the event processor to shut down after 5 seconds; some analytics events may not have been delivered
RESULT: stop() returned after 5.0s -- BOUNDED (fixed)
```

Two tests added:

- `test_stop_returns_even_if_event_delivery_never_completes` — the invariant. Stalls a flush worker on a payload that never completes and asserts `stop()` still returns. I verified this fails against the unfixed code (`stop() never returned; it is waiting on a delivery that never completes`) rather than only passing against the fix.
- `test_stop_still_delivers_buffered_events` — a control, so that a `stop()` which simply gave up immediately could not pass the first test. It asserts buffered events are still delivered when delivery works normally.

`LD_SKIP_DATABASE_TESTS=1 uv run pytest`: 1417 passed, 279 skipped. `mypy`, `isort --check`, and `pycodestyle` all clean.

## Not addressed here

`DefaultAsyncEventProcessor.stop()` has the same unbounded shape (`async_event_processor.py:245`), while `flush_and_wait()` directly above it already bounds its wait with `asyncio.wait_for`. I have not reproduced a hang on that path, so I left it out rather than change code I could not verify — but it looks worth a follow-up.

---

> [!NOTE]
> **Overview**
> Adds **`shutdown_timeout`** on `Config` (default **5 seconds**) so `LDClient.close()` / event processor `stop()` no longer wait forever when final analytics delivery stalls—e.g. **hung DNS** outside `HTTPConfig` connect/read timeouts.
> 
> Shutdown now shares one deadline via **`_Deadline`**: **`FixedThreadPool.wait(timeout)`** returns whether workers drained; **`_post_message_and_wait`** bounds inbox **`put`** and reply **`wait`**; **`EventDispatcher._do_shutdown`** applies the same budget to flush and diagnostic pools. On timeout the SDK logs a warning and returns; undelivered events may be dropped. **`shutdown_timeout=None`** restores indefinite wait.
> 
> Tests cover **`stop()`** returning while delivery never completes and still flushing buffered events when delivery works.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f7d6c8d30b4fb2e687841e5cbf7af77b0e83c503. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/config.py
+++ b/ldclient/config.py
@@ -321,6 +321,7 @@ class Config(DataSourceBuilderConfig, PrivateAttributesConfig):
         omit_anonymous_contexts: bool = False,
         payload_filter_key: Optional[str] = None,
         datasystem_config: Optional[DataSystemConfig] = None,
+        shutdown_timeout: Optional[float] = 5,
     ):
         """
         :param sdk_key: The SDK key for your LaunchDarkly account. This is always required.
@@ -392,6 +393,12 @@ class Config(DataSourceBuilderConfig, PrivateAttributesConfig):
         :param omit_anonymous_contexts: Sets whether anonymous contexts should be omitted from index and identify events.
         :param payload_filter_key: The payload filter is used to selectively limited the flags and segments delivered in the data source payload.
         :param datasystem_config: Configuration for the upcoming enhanced data system design. This is experimental and should not be set without direction from LaunchDarkly support.
+        :param shutdown_timeout: The maximum number of seconds that :func:`ldclient.client.LDClient.close()`
+          will wait for pending analytics events to be delivered before giving up. Delivery of an event
+          payload can block for an unbounded time when name resolution hangs, because DNS lookups are not
+          covered by the connect and read timeouts in :class:`HTTPConfig`; without this limit, ``close()``
+          would never return. Any events still undelivered when the timeout expires are discarded. Set this
+          to ``None`` to wait indefinitely, but be aware that doing so can prevent your process from exiting.
         """
         self.__sdk_key = validate_sdk_key_format(sdk_key, log)
 
@@ -429,6 +436,7 @@ class Config(DataSourceBuilderConfig, PrivateAttributesConfig):
         self.__enable_event_compression = enable_event_compression
         self.__omit_anonymous_contexts = omit_anonymous_contexts
         self.__payload_filter_key = payload_filter_key
+        self.__shutdown_timeout = None if shutdown_timeout is None else max(shutdown_timeout, 0)
         self._data_source_update_sink: Optional[DataSourceUpdateSink] = None
         self._instance_id: Optional[str] = None
         self._datasystem_config = datasystem_config
@@ -643,6 +651,14 @@ class Config(DataSourceBuilderConfig, PrivateAttributesConfig):
         Determines whether or not anonymous contexts will be omitted from index and identify events.
         """
         return self.__omit_anonymous_contexts
+
+    @property
+    def shutdown_timeout(self) -> Optional[float]:
+        """
+        The maximum number of seconds that :func:`ldclient.client.LDClient.close()` will wait for
+        pending analytics events to be delivered before giving up, or ``None`` to wait indefinitely.
+        """
+        return self.__shutdown_timeout
 
     @property
     def payload_filter_key(self) -> Optional[str]:
--- a/ldclient/impl/events/event_processor.py
+++ b/ldclient/impl/events/event_processor.py
@@ -10,6 +10,7 @@ import uuid
 from collections import namedtuple
 from random import Random
 from threading import Event, Lock, Thread
+from typing import Optional
 
 import urllib3
 
@@ -38,6 +39,21 @@ __MAX_FLUSH_THREADS__ = 5
 
 
 EventProcessorMessage = namedtuple('EventProcessorMessage', ['type', 'param'])
+
+
+class _Deadline:
+    """
+    Tracks how much of a time budget is left, so that a sequence of blocking waits can share a
+    single overall limit. A timeout of None means there is no limit.
+    """
+
+    def __init__(self, timeout: Optional[float]):
+        self._end = None if timeout is None else time.monotonic() + timeout
+
+    def remaining(self) -> Optional[float]:
+        if self._end is None:
+            return None
+        return max(0.0, self._end - time.monotonic())
 
 
 class EventPayloadSendTask:
@@ -161,12 +177,21 @@ class EventDispatcher(EventDispatcherBase):
             self._diagnostic_flush_workers.execute(task.run)
 
     def _do_shutdown(self):
+        # Delivery of an event payload can block for an unbounded time - notably when name
+        # resolution hangs, which the connect and read timeouts do not cover - so these waits are
+        # bounded. Worker threads are daemons, so any that are still stuck do not keep the process
+        # alive; the events they were carrying are simply lost.
+        deadline = _Deadline(self._config.shutdown_timeout)
+
         self._flush_workers.stop()
-        self._flush_workers.wait()
+        drained = self._flush_workers.wait(deadline.remaining())
 
         if self._diagnostic_flush_workers:
             self._diagnostic_flush_workers.stop()
-            self._diagnostic_flush_workers.wait()
+            drained = self._diagnostic_flush_workers.wait(deadline.remaining()) and drained
+
+        if not drained:
+            log.warning("Timed out waiting for analytics events to be delivered while shutting down; some events were dropped")
 
         if self._close_http:
             self._http.clear()
@@ -188,6 +213,7 @@ class DefaultEventProcessor(EventProcessor):
 
         self._close_lock = Lock()
         self._closed = False
+        self._shutdown_timeout = config.shutdown_timeout
 
         (dispatcher_class or EventDispatcher)(self._inbox, config, http, diagnostic_accumulator)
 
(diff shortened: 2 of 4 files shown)


jsonbailey:

Issue #495, opened 2026-08-15
LDClient.close() leaks components when one fails to stop, and is not idempotent


aviadr1:
## Describe the bug

`LDClient.close()` is documented as "Releases all threads and network connections used by the LaunchDarkly client", but it cannot honour that in two situations.

Found while investigating #493 — a production incident where `close()` hung and left worker pods alive for hours to days, each holding a concurrency slot. These two are separate defects in the same method, found by auditing the shutdown path afterwards. I have not observed either in our own production, but both reproduce reliably.

### 1. A failure in one component orphans the rest

`ldclient/client.py:355-358`:

```python
self._event_processor.stop()
self._data_system.stop()
self.__big_segment_store_manager.stop()
```

There is no error handling. If the first raises, the data system and the big segment store manager are never stopped.

This is not contrived: two of these three reach code the SDK does not control — `_data_system.stop()` reaches the eventsource client, and `__big_segment_store_manager.stop()` calls `stop()` on the **application's own `BigSegmentStore` implementation**. A third-party store raising on shutdown is entirely ordinary. The caller is then left with leaked threads and connections belonging to a client it believes is closed.

### 2. `close()` is not idempotent

`LDClient` has no closed-flag, so a second `close()` runs the whole sequence again. That calls `stop()` a second time on the application's `BigSegmentStore`, and under FDv2 calls `store.close()` twice. `DefaultEventProcessor.stop()` is the only layer that self-guards.

Double-close is easy to reach by accident — an explicit `close()` inside a `with` block does it, as does any cleanup path that can run twice.

## To reproduce

PR #498 adds two failing tests (marked `xfail(strict=True)` so CI stays green). Run them with `--runxfail` to see the real failures.

For the idempotency case the recorded shutdown log is:

```
['update_processor', 'big_segment_store', 'update_processor', 'big_segment_store']
```

## Expected behavior

1. `close()` releases every component even if one fails — log the failure and continue.
2. `close()` on an already-closed client does nothing further.

## SDK version

9.16.1 / `main` at 5da1515

## Language version, developer tools

Python 3.14

## OS/platform

Linux

## Additional context

I have deliberately not proposed a fix, since the error-propagation semantics (whether `close()` should still raise after releasing everything) is a design call for the SDK team. The tests pin the leak, not the exception behaviour.

aviadr1:
One more way to reach §2 (non-idempotency), from production use.

[...]

PR #498, opened 2026-08-15
test: Add failing tests for LDClient.close() releasing all components

aviadr1:
Failing tests for #495. **Tests only — no fix.**

Found while investigating #493, a production incident where `LDClient.close()` hung and left worker pods alive for hours to days, each holding a concurrency slot. Auditing the shutdown path afterwards turned up two further defects in `close()` itself. I have not observed these two in our own production, but both reproduce reliably.

## What the tests pin

`close()` is documented as "Releases all threads and network connections used by the LaunchDarkly client". Two ways it does not:

**`test_close_releases_every_component_even_if_one_raises`** — `client.py:355-358` calls the event processor, data system and big segment store manager in sequence with no error handling. If the first raises, the other two are never stopped. Two of the three reach code the SDK does not control: the eventsource client, and the application's own `BigSegmentStore` implementation. The caller is left with leaked threads and connections from a client it believes is closed.

**`test_close_is_idempotent`** — there is no closed-flag, so a second `close()` re-runs the whole sequence. The recorded shutdown log is `['update_processor', 'big_segment_store', 'update_processor', 'big_segment_store']` — the application's own store is stopped twice, and under FDv2 `store.close()` is called twice.

## Why xfail

Both are marked `@pytest.mark.xfail(strict=True)` so CI stays green while the defect is documented in the suite. Because it is `strict`, the moment the behaviour is fixed these turn into failures telling you to drop the marker.

To see the actual failures:

```
uv run pytest ldclient/testing/test_ldclient_shutdown.py --runxfail
```

## Why no fix

Whether `close()` should still raise after releasing everything is a semantics decision for the SDK team, so the tests pin the leak rather than the exception behaviour. Happy to add the fix if you tell me which way you want it.

---

> [!NOTE]
> **Overview**
> Adds **`ldclient/testing/test_ldclient_shutdown.py`** — contract tests for sync **`LDClient.close()`** with **no production code changes**. Both cases are **`@pytest.mark.xfail(strict=True)`** so CI stays green until a fix lands.
> 
> **`test_close_releases_every_component_even_if_one_raises`** documents that **`client.py`** stops the event processor, data system, and big-segment manager in sequence with no error handling; if **`stop()`** on the first raises (e.g. eventsource or user **`BigSegmentStore`**), the others never run.
> 
> **`test_close_is_idempotent`** documents that without a closed flag, a second **`close()`** runs shutdown again (e.g. double **`stop()`** on the app’s big-segment store).
> 
> Uses injectable **`Recording*`** / **`FailingEventProcessor`** helpers and unreachable URIs so tests don’t need network.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a10751c1d65bfb0978f66f7b14ab190887d5ce03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/testing/test_ldclient_shutdown.py
+++ b/ldclient/testing/test_ldclient_shutdown.py
@@ -1,0 +1,142 @@
+"""
+Tests for the shutdown contract of LDClient.close().
+
+close() is documented as "Releases all threads and network connections used by the LaunchDarkly
+client". These tests pin that contract. They are currently marked xfail because the SDK does not
+yet honour it; each one describes a specific way the contract is broken.
+"""
+
+import time
+
+import pytest
+
+from ldclient.client import Config, LDClient
+from ldclient.config import BigSegmentsConfig
+from ldclient.interfaces import (
+    BigSegmentStore,
+    BigSegmentStoreMetadata,
+    EventProcessor,
+    UpdateProcessor
+)
+
+unreachable_uri = "http://fake"
+
+
+class RecordingBigSegmentStore(BigSegmentStore):
+    """A user-supplied big segment store, of the kind an application would provide."""
+
+    def __init__(self, log: list):
+        self._log = log
+
+    def get_metadata(self) -> BigSegmentStoreMetadata:
+        return BigSegmentStoreMetadata(int(time.time() * 1000))
+
+    def get_membership(self, user_hash: str):
+        return None
+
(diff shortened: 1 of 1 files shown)


jsonbailey:

Issue #496, opened 2026-08-15
set_config() blocks every concurrent ldclient.get() while closing the previous client


aviadr1:
## Describe the bug

`ldclient.set_config()` holds the global write lock across both the construction of the replacement client **and** `old_client.close()`, so every concurrent `ldclient.get()` blocks for the entire duration.

`ldclient/__init__.py:38-47`:

```python
with __lock.write():
    try:
        if __client:
            new_client = LDClient(config=config, start_wait=start_wait)   # blocks up to start_wait
            old_client = __client
            __client = new_client
            old_client.close()                                           # blocks on network shutdown
    finally:
        __config = config
```

`ReadWriteLock.lock()` (`ldclient/impl/rwlock.py`) acquires the underlying mutex and holds it for the whole write section, and `rlock()` must acquire that same mutex. So while `set_config()` runs, every `ldclient.get()` — which is on the hot path of every flag evaluation in an application using the singleton API — is blocked.

This couples a network-dependent shutdown to a lock that every evaluating thread needs. A slow `close()` stalls the whole application, not just the thread that called `set_config()`.

Found while investigating #493, a production incident where `close()` hung and left worker pods alive for hours to days. This is what makes that class of hang so damaging under the singleton API: before a shutdown timeout existed, a stalled `close()` inside `set_config()` would block every `get()` in the process indefinitely. I have not observed this specific path in our own production — we do not call `set_config()` after startup — but the coupling is real and reproduces reliably.

Even with a bounded `close()`, the window is `start_wait` (default 5s) plus the shutdown timeout, so a reconfigure can stall all evaluating threads for ~10 seconds.

## To reproduce

PR #499 adds a failing test (marked `xfail(strict=True)`; run with `--runxfail` to see it fail). It drives a 3 second `close()` and measures a concurrent `get()`:

```
AssertionError: ldclient.get() blocked for 2.8s while set_config() was closing the previous client
```

## Expected behavior

Reconfiguring should not stall threads that are only reading the shared client. The write lock should cover the pointer swap, not the construction and teardown around it.

## SDK version

9.16.1 / `main` at 5da1515

## Language version, developer tools

Python 3.14

## OS/platform

Linux

## Additional context

I have not proposed a fix. Moving construction and `close()` outside the lock is the obvious direction, but doing it without introducing a lost-update race between concurrent `set_config()` callers is a design decision I would rather leave to you.

aviadr1:
[...]

PR #499, opened 2026-08-15
test: Add failing test for set_config() blocking concurrent get()

aviadr1:
Failing test for #496. **Test only — no fix.**

Found while investigating #493, a production incident where `LDClient.close()` hung and left worker pods alive for hours to days. This issue is what makes that class of hang so damaging under the singleton API. I have not observed this specific path in our own production — we do not call `set_config()` after startup — but the coupling is real and reproduces reliably.

## What the test pins

`set_config()` holds the global write lock across both the construction of the replacement client and `old_client.close()` (`ldclient/__init__.py:38-47`). `ReadWriteLock.lock()` holds the underlying mutex for the whole write section and `rlock()` needs that same mutex, so every concurrent `ldclient.get()` blocks — and `get()` is on the hot path of every flag evaluation in an application using the singleton API.

`test_set_config_does_not_block_concurrent_get` drives a 3 second `close()` and measures a concurrent `get()`:

```
AssertionError: ldclient.get() blocked for 2.8s while set_config() was closing the previous client
```

Before a shutdown timeout existed, an unbounded `close()` here would block every `get()` in the process indefinitely. Even bounded, the window is `start_wait` (default 5s) plus the shutdown timeout, so a reconfigure can stall all evaluating threads for roughly ten seconds.

## Why xfail

Marked `@pytest.mark.xfail(strict=True)` so CI stays green while the defect is documented; being strict, it fails loudly once fixed so the marker gets removed. To see the real failure:

```
uv run pytest ldclient/testing/test_ldclient_singleton.py -k set_config_does_not_block --runxfail
```

## Why no fix

Moving construction and `close()` outside the lock is the obvious direction, but doing it without introducing a lost-update race between concurrent `set_config()` callers is a design decision I would rather leave with you.

---

> [!NOTE]
> **Overview**
> **Test-only change** for issue #496: documents a singleton API defect where **`set_config()` keeps the global write lock** while it builds the replacement client and runs **`old_client.close()`**, so concurrent **`ldclient.get()`** (read lock) stalls on the same mutex.
> 
> Adds **`test_set_config_does_not_block_concurrent_get`**, which patches **`LDClient.close`** to sleep 3s during reconfigure and asserts a concurrent **`get()`** finishes in under half that time. The test **fails today** (~2.8s block) and is marked **`@pytest.mark.xfail(strict=True)`** so CI stays green until a fix lands.
> 
> No library changes; the PR pins the invariant that reconfigure should not couple slow shutdown to the flag-evaluation hot path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce3e2722eedbc552712def2a7429f10dd03c17f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient/testing/test_ldclient_singleton.py
+++ b/ldclient/testing/test_ldclient_singleton.py
@@ -1,7 +1,12 @@
 import json
+import threading
+import time
+
+import pytest
 
 import ldclient
 from ldclient import _reset_client
+from ldclient.client import LDClient
 from ldclient.config import Config
 from ldclient.testing.http_util import BasicResponse, start_server
 from ldclient.testing.stub_util import make_put_event, stream_content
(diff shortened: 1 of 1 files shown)


jsonbailey:

PR #519, opened 2026-09-10
feat: Remove permanent failure modes from FDv1 following RETRY spec

jsonbailey:
BEGIN_COMMIT_OVERRIDE
feat: Retry indefinitely after a data source failure instead of stopping permanently
fix: Warn and use the documented default for an invalid poll interval or initial reconnect delay
END_COMMIT_OVERRIDE

## Summary

Brings the FDv1 streaming and polling data sources into conformance with the [RETRY specification](https://github.com/launchdarkly/sdk-specs/tree/main/specs/RETRY-retry-behavior). No HTTP response and no transport-level failure stops a data source permanently any more.

Every failure is classified `normal` or `unexpected`. A normal failure retries on the existing curve — 1s doubling to a 30s ceiling for streaming, the poll interval for polling. An unexpected failure (`401`, `403`, any other `4xx` outside `400`/`408`/`429`) moves to a longer regime starting at 5 minutes and doubling to a 1-hour ceiling, and keeps retrying there until the condition clears.

The retry state machine itself landed in #522; the delay-source plumbing landed in #521. This PR connects them to the four data sources and removes the permanent-stop paths.

## Behaviour changes for release notes

1. **A bad or revoked SDK key no longer fails fast.** `LDClient(config, start_wait=N)`, `postfork(start_wait=N)` and `await client.start(start_wait=N)` now block for the full `start_wait` and return with `is_initialized()` false, rather than returning at once. The SDK keeps retrying in the background.
2. **`DataSourceState.OFF` is now reserved for explicit shutdown and unparseable configuration.** HTTP errors produce `INTERRUPTED`.
3. **A rejected SDK key now logs at `error` roughly hourly, indefinitely**, rather than once. An SDK retrying with a rejected credential consumes resources, and the condition needs a person to fix it.
4. **A server-initiated stream close now logs a warning** where it previously logged nothing, and backs off rather than reconnecting immediately.
5. **Failure logs now state the actual delay** — `Received HTTP error 401 (invalid SDK key) for stream connection - will retry in 300.0s`. Previously the SDK said only "will retry", and the delay was logged separately at `info` by the SSE client, so it was invisible at default log levels.
6. **An invalid `poll_interval` or `initial_reconnect_delay` now logs a warning and uses the documented default.** `poll_interval` was silently clamped and `initial_reconnect_delay` was not checked at all.

## What changed

- **`impl/datasource/{streaming,async_streaming,polling,async_polling}.py`** — the permanent-stop paths are gone. Each failure is classified, the retry state advances, status becomes `INTERRUPTED`, and the wait is interruptible by `stop()`, which matters now that a wait can be an hour long.
- **`config.py` / `async_config.py`** — both intervals are validated. `Config` previously clamped `poll_interval` with `max()`, which let `NaN` and `inf` through, because every comparison against `NaN` is false. A `NaN` interval reached `Event.wait()` and the delay arithmetic. The 30-second minimum still applies on top of validation.
- **`impl/util.py`** — `validate_positive_finite`, beside the validators `Config` already imports, so `config` and `retry` share it without either importing the other.
- **`impl/retry.py`** — uses the shared validator; the two configurable defaults now live in `config.py` next to `DEFAULT_STREAM_URI`.
- **`impl/datasource/datasource_common.py`**, **`interfaces.py`**, **`client.py`** / **`async_client.py`** — docstrings and status handling updated for the above.
- **`impl/aio/transport.py`** — the async transport's own retry is driven by the shared state.
- **`contract-tests/`** — both services declare the conformance capability.

## Known gaps, deliberately out of scope

- **FDv2** (`impl/datasourcev2/**`, `impl/datasystem/**`) still stops permanently on an unexpected response. Tracked as SDK-2776.
- **The event processor's `_disabled` permanent stop** stays. No spec binding exists for it yet. Also SDK-2776.
- **SSE action loops have no `except`.** "A data source never stops" holds via `ld_eventsource` internals rather than by construction. No reachable escape path was found at the pinned version.

## Testing

`make test`: 1681 passed. `make lint`: clean across 228 files.

Contract tests were run out of band against harness v2.41.0 — streaming's eight conformance subtests and polling's four all pass. Note the conformance scenarios need `-enable-long-running-tests`, which the `Makefile` does not pass, so they do not run in CI.

---

> [!NOTE]
> **Overview**
> Aligns **FDv1 streaming and polling** (sync and async) with the RETRY spec: data source failures no longer shut down permanently. Failures are classified as **normal** vs **unexpected** (e.g. most 4xx including invalid SDK keys), drive shared **`RetryState`** backoff, report **`INTERRUPTED`**, and keep retrying—including multi-minute/hour waits that **`stop()`** can interrupt.
> 
> **Retry ownership moves to the SDK**: SSE clients are created with **`sdk_managed_retry`** so reconnect delays come from **`for_streaming` / `for_polling`**, not the eventsource library. Server-initiated stream closes are treated as **`StreamClosedError`** with normal backoff (not ignored tight reconnect loops).
> 
> **User-visible behavior shifts**: bad SDK keys no longer fail fast during **`start_wait`**—the client returns uninitialized but keeps retrying in the background. **`DataSourceState.OFF`** is reserved for explicit shutdown (not HTTP errors). Logs now include the **actual retry delay**; invalid **`poll_interval`** / **`initial_reconnect_delay`** warn and fall back to shared defaults in **`config.py`**.
> 
> Contract test services advertise **`retry-conformance-fdv1-*`** capabilities; docs and tests are updated across processors, transport, and e2e cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30bb8ecc32e57a694be4fdcbafd7265f3bc6bbb3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ldclient/impl/aio/concurrency.py
```diff
@@ -189,33 +189,41 @@ async def stop_all(self, timeout: float = 1) -> None:
...
     Mirrors the semantics of ``ldclient.impl.repeating_task.RepeatingTask``:
-    the interval is measured from the start of each invocation, exceptions
-    from the callback are logged, and ``stop()`` prevents any further
-    invocations but cannot be undone."""
+    the wait starts when the callback returns, exceptions from the callback
+    are logged, and ``stop()`` prevents any further invocations but cannot be
+    undone."""
 
-    def __init__(self, label: str, interval: float, initial_delay: float, callable: Callable):
+    def __init__(self, label: str, delays: DelaySource, initial_delay: float, callable: Callable):
```
jsonbailey: Can we include the initial delay with the delay source?

ldclient/impl/aio/transport.py
```diff
@@ -118,11 +118,16 @@ def __init__(self, config, session: Optional[aiohttp.ClientSession] = None, prox
         self._http_options = http_options if http_options is not None else config.http
         self._proxy = proxy if proxy is not None else (self._http_options.http_proxy or None)
 
-    def create(self, url: str, initial_retry_delay: float, query_params=None) -> AsyncSSEClient:
-        """Builds an SSE client for the given stream URL. Headers, timeouts,
-        proxy settings, and the retry/backoff policy come from the SDK config.
-        ``query_params`` is an optional zero-argument callable evaluated on
-        each (re)connect to produce additional query string parameters."""
+    def create(self, url: str, initial_retry_delay: float, query_params=None, sdk_managed_retry: bool = False) -> AsyncSSEClient:
```
jsonbailey: This is temporary until FDv2 is updated.

ldclient/impl/datasource/async_streaming.py
```diff
@@ -51,14 +61,18 @@ def __init__(self, config, store, ready, diagnostic_accumulator, sse_factory: Op
...
+        self._start_requested = False
+        self._retry = retry_state or for_streaming(config.initial_reconnect_delay)
+        self._signalled_healthy = False
+        self._interrupted_by_sdk = False
 
     def start(self):
-        if self._started:
-            raise RuntimeError("processors can only be started once")
-        self._started = True
+        if self._start_requested:
```
jsonbailey: Lets keep the name as _started.

ldclient/impl/datasource/async_streaming.py
```diff
@@ -156,7 +182,28 @@ async def __stop_with_error_info(self, error: Optional[DataSourceErrorInfo]):
...
+
+        It fires once per stream. A later message on the same stream must not
+        restart the reset window. The SSE client's own signal is no use here
+        because it fires when the connection opens, and an open connection that
+        has sent no data yet does not show the stream is working.
+        """
+        if self._signalled_healthy:
+            return
+        self._signalled_healthy = True
+        self._retry.record_healthy()
```
jsonbailey: It seems like we can send this unconditionally and remove the signalled_healthy check from this class. Let's double check this.

ldclient/impl/retry.py
```diff
@@ -0,0 +1,370 @@
...
+    """Resets once the component has operated without failing for
+    ``seconds``. This is the streaming policy."""
+
+    def __init__(self, seconds: float):
+        self._seconds = seconds
+        self._healthy_since: Optional[float] = None
+
+    def note_healthy(self) -> None:
+        if self._healthy_since is None:
+            self._healthy_since = time.time()
```
jsonbailey: Look at using time.monotonic()

ldclient/impl/retry.py
```diff
@@ -0,0 +1,370 @@
...
+        """Reports whether the reset condition is met."""
+        ...
+
+
+class AfterHealthyFor(ResetPolicy):
+    """Resets once the component has operated without failing for
+    ``seconds``. This is the streaming policy."""
+
+    def __init__(self, seconds: float):
+        self._seconds = seconds
```
jsonbailey: This should have a better name, seconds for what?

ldclient/impl/retry.py
```diff
@@ -0,0 +1,370 @@
...
+            self._min_delay = self._extended_initial_delay
+            self._max_delay = max(self._extended_ceiling, self._min_delay)
+            self._n = 1
+        else:
+            self._n += 1
+
+        self._next_delay = self._compute_wait(wait_override)
+        return self._next_delay
+
+    def record_success(self) -> float:
```
jsonbailey: Can we combine record_success and record_healthy into one method if possible to remove ambiguity on which to call.

ldclient/impl/datasource/async_streaming.py
```diff
@@ -141,12 +166,10 @@ def _record_stream_init(self, failed: bool):
 
     async def stop(self):
```
jsonbailey: Lets double check that when we call stop, the sleep delay inside handle_error could cause the stop to take a long time to process.

ldclient/impl/aio/transport.py
```diff
@@ -134,6 +139,23 @@ def create(self, url: str, initial_retry_delay: float, query_params=None) -> Asy
         proxy = self._proxy or _get_proxy_url(url)
         if proxy:
             aiohttp_request_options["proxy"] = proxy
+        if sdk_managed_retry:
+            # The SSE client's retry is disabled; the SDK owns the delay.
+            retry_options: dict = {
+                "initial_retry_delay": 0,
+                "retry_delay_strategy": RetryDelayStrategy(),
```
tanderson-ld, line 146: Is `RetryDelayStrategy()` an unused placeholder in this case?

ldclient/impl/datasource/async_polling.py
```diff
@@ -43,7 +57,12 @@ def initialized(self):
...
 
     async def stop(self):
-        self.__stop_with_error_info(None)
+        log.info("Stopping AsyncPollingUpdateProcessor")
+        self._task.stop()
+
+        if self._data_source_update_sink is not None:
+            self._data_source_update_sink.update_status(DataSourceState.OFF, None)
+
         # Wait for the current poll to finish before closing the transport, so we do
```
tanderson-ld, line 66: Off is reported, but then the comment `# Wait for the current poll to finish before closing the transport` makes it seem like it is still doing things?

ldclient/impl/datasource/async_streaming.py
```diff
@@ -198,46 +226,54 @@ async def _process_message(self, msg: Event) -> bool:
...
             error_info = DataSourceErrorInfo(DataSourceErrorKind.ERROR_RESPONSE, error.status, time.time(), str(error))
+            description = "Received %s for stream connection" % http_error_description(error.status)
+            level = log.error if kind is FailureKind.UNEXPECTED else log.warning
+        elif isinstance(error, StreamClosedError):
+            kind = FailureKind.NORMAL
+            error_info = DataSourceErrorInfo(DataSourceErrorKind.NETWORK_ERROR, 0, time.time(), str(error))
+            description = "The server closed the stream connection"
+            level = log.warning
+        else:
+            # A certificate failure lands here too, and is as normal as the rest.
```
tanderson-ld, line 259:
```suggestion
```

ldclient/impl/datasource/async_streaming.py
```diff
@@ -198,46 +226,54 @@ async def _process_message(self, msg: Event) -> bool:
...
-            self._connection_attempt_start_time = None
+        if self._data_source_update_sink is not None:
+            self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, error_info)
 
-            if self._data_source_update_sink is not None:
-                self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, DataSourceErrorInfo(DataSourceErrorKind.UNKNOWN, 0, time.time(), str(error)))
-            # no stacktrace here because, for a typical connection error, it'll just be a lengthy tour of HTTP client internals
-        self._connection_attempt_start_time = time.time() + self._sse.next_retry_delay
-        return True
+        self._connection_attempt_start_time = time.time() + delay
```
tanderson-ld, line 273: Probably out of scope since it is likely existing, but this _connection_attempt_start_time being calculated before the sleep means clock shift during the sleep will not self correct.  The delay mechanism is safe, but the diagnostic value will be incorrect.

ldclient/impl/datasource/polling.py
```diff
@@ -54,46 +67,54 @@ def initialized(self):
...
-                    self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, error_info)
+            description = "Received %s for polling request" % http_error_description(e.status)
+            level = log.error if kind is FailureKind.UNEXPECTED else log.warning
+            stacktrace = None
         except Exception as e:
-            log.exception('Error: Exception encountered when updating flags. %s' % e)
-
-            if self._data_source_update_sink is not None:
-                self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, DataSourceErrorInfo(DataSourceErrorKind.UNKNOWN, 0, time.time(), str(e)))
+            # A certificate failure lands here too, and is as normal as the rest.
```
tanderson-ld, line 103:
```suggestion
```

ldclient/impl/datasource/streaming.py
```diff
@@ -94,9 +108,14 @@ def run(self):
                         log.info("StreamingUpdateProcessor initialized ok.")
                         self._ready.set()
             elif isinstance(action, Fault):
-                # If the SSE client detects the stream has closed, then it will emit a fault with no-error. We can
-                # ignore this since we want the connection to continue.
+                # A Fault with no error is a clean close. An interrupt the SDK
+                # asked for is not a failure.
                 if action.error is None:
+                    if self._interrupted_by_sdk:
+                        self._interrupted_by_sdk = False
```
tanderson-ld, line 115: A boolean like this being reused worries me a tad, especially since the bool is on self and not on the particular connection instance.

ldclient/impl/datasource/streaming.py
```diff
@@ -176,46 +202,50 @@ def _process_message(self, store, msg: Event) -> bool:
...
             error_info = DataSourceErrorInfo(DataSourceErrorKind.ERROR_RESPONSE, error.status, time.time(), str(error))
+            description = "Received %s for stream connection" % http_error_description(error.status)
+            level = log.error if kind is FailureKind.UNEXPECTED else log.warning
+        elif isinstance(error, StreamClosedError):
+            kind = FailureKind.NORMAL
+            error_info = DataSourceErrorInfo(DataSourceErrorKind.NETWORK_ERROR, 0, time.time(), str(error))
+            description = "The server closed the stream connection"
+            level = log.warning
+        else:
+            # A certificate failure lands here too, and is as normal as the rest.
```
tanderson-ld, line 233:
```suggestion
```

ldclient/impl/datasource/streaming.py
```diff
@@ -176,46 +202,50 @@ def _process_message(self, store, msg: Event) -> bool:
...
-            self._connection_attempt_start_time = None
+        if self._data_source_update_sink is not None:
+            self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, error_info)
 
-            if self._data_source_update_sink is not None:
-                self._data_source_update_sink.update_status(DataSourceState.INTERRUPTED, DataSourceErrorInfo(DataSourceErrorKind.UNKNOWN, 0, time.time(), str(error)))
-            # no stacktrace here because, for a typical connection error, it'll just be a lengthy tour of urllib3 internals
-        self._connection_attempt_start_time = time.time() + self._sse.next_retry_delay
-        return True
+        self._connection_attempt_start_time = time.time() + delay
```
tanderson-ld, line 247: Same clock issue here, but preexisting.  Up to you if you want to address it.

ldclient/impl/datasource/polling.py
tanderson-ld, line 1: There is an existing issue where OFF->VALID can happen due to pending request/response coming in.  I think this should be fixed as part of this work since the hood is open.  The SDKs I had worked on also had some of these situations and I addressed them by encapsulating the status reporting in "try set status" where the first Off latches a bool and future statuses are ignored.

I think this also indicates the request/response routine isn't interruptible, but that seems less important.

ldclient/impl/datasource/streaming.py
tanderson-ld, line 1: There is an existing issue where OFF->INTERRUPTED can happen.  Same "try set status" with latching Off should work here too.

tanderson-ld, line 75: AI is flagging a case where it is possible for a stream to start after close and stay connected indefinitely.  This is also related to a race where after close() is called, it can get connected and report Off -> Valid and put data in the data store.  This is existing, but seems like it should probably be addressed with this work.

tanderson-ld (changes requested):

ldclient/impl/aio/transport.py
jsonbailey: No, I added a comment but the base RetryDelayStrategy has a retry of 0 seconds effectively disabling the retry.

ldclient/impl/datasource/polling.py
jsonbailey: This should be addressed

ldclient/impl/datasource/streaming.py
jsonbailey: Should be fixed.

tanderson-ld (approved):

PR #522, opened 2026-09-14
feat: Add retry state for RETRY-spec backoff

jsonbailey:
## Summary

Adds the retry state machine that the FDv1 data sources will use for RETRY-spec conformance. Pure addition — nothing imports it yet, so it can be reviewed against the spec without reading any data-source code.

`RetryState` answers one question: how long to wait before the next attempt. It serves all four data sources; streaming and polling differ only in their construction parameters and their reset policy.

## The algorithm

Every failure is classified `NORMAL` or `UNEXPECTED`. `400`, `408`, `429` and all `5xx` are normal; every other `4xx`, including `401` and `403`, is unexpected. Only an HTTP status can be unexpected — every network and TLS failure is normal.

A normal failure advances the delay on the current curve. An unexpected failure raises both bounds and keeps them raised until the reset condition is met, so a normal failure that follows cannot lower them.

```
streaming, normal:    1 -> 2 -> 4 -> 8 -> 16 -> 30 -> 30 ...
streaming, extended:  300 -> 600 -> 1200 -> 2400 -> 3600 -> 3600 ...
polling, normal:      poll_interval, flat — no escalation
polling, extended:    max(300, poll_interval) doubling to max(3600, poll_interval)
```

Each delay is then reduced by a jitter of up to half itself, and never falls below the caller's operating cadence — `poll_interval` for polling, absent for streaming.

## Two things that are easy to get wrong

**A backoff wait applies to a retry, not to every operation.** `record_success` restores the operating cadence even while the retry state is still raised, so a recovered service is polled at its normal rate immediately. Another SDK shipped without this: after an outage its first successful poll still waited twenty minutes.

**Clearing the retry state is separate from restoring the cadence.** `record_success` does the latter; `_reset_if_due` does the former, and only when the reset policy is satisfied. Streaming's policy is 60 seconds of continuous healthy operation, polling's is two consecutive successes.

`_reset_if_due` is called from `record_failure` because nothing runs while a stream is healthy — the moment the 60-second threshold is crossed is otherwise unobservable, so the next failure is the only place it can be noticed.

## Why TLS failures are normal

The narrow classification was built first and reverted. All SDK traffic is HTTPS, so a genuine certificate problem cannot be distinguished from an ordinary transient fault across platforms — a peer sending FIN mid-handshake surfaces differently from one sending RST, and the classification would depend on which an intermediary happened to send. A connection flapping faster than the reset threshold then ratchets to the hour ceiling with no way out.

## Testing

`make test`: 1599 passed. `make lint`: clean across 226 files.

71 tests covering both delay ladders exactly, jitter bounds, ceiling stickiness, both reset policies, the cadence floor, and the classification table. No test sleeps: the clock and the jitter are patched at the module level, so the 60-second reset and a 20-cycle flapping scenario both run instantly.

---

> [!NOTE]
> **Overview**
> Introduces a standalone **`RetryState`** module (`ldclient/impl/retry.py`) that will drive FDv1 data-source reconnect/poll timing. Nothing in production imports it yet.
> 
> Failures are classified as **normal** vs **unexpected** (via `classify_http_status`: most 4xx except 400/408/429 are unexpected; 5xx and transport errors stay normal). Backoff **doubles to a ceiling**, subtracts up to **half jitter**, and never waits less than the **operating cadence** (zero for streaming, `poll_interval` for polling). An unexpected failure switches to an **extended** delay ladder (5–60 minutes) that stays sticky until reset.
> 
> **Streaming** uses `for_streaming` with exponential normal delays (default 1s → 30s cap) and resets after **60s** of healthy operation. **Polling** uses `for_polling` with flat normal retries at the poll interval and resets after **two consecutive successes**. **`record_success`** restores the healthy cadence immediately even while extended state is still raised; clearing extended state is separate and gated by the reset policy.
> 
> Adds **`test_retry.py`** (delay tables, jitter bounds, reset policies, invalid config fallbacks) plus **`test_util`** helpers to patch retry’s `time`/`random` without sleeping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d300c42bcc4ec616a6cee9625df752bf121ad42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+from ldclient.impl.util import log
+
+# The delay bounds of the extended regime, in seconds. A component enters the
+# extended regime after an unexpected failure.
+EXTENDED_INITIAL_DELAY = 5 * 60
+EXTENDED_MAX_DELAY = 60 * 60
+
+# The delay bounds of the normal regime for streaming, in seconds. The initial
+# delay is configurable as ``initial_reconnect_delay``.
+STREAMING_MAX_DELAY = 30
```
tanderson-ld: Should this distinguish itself from extended? Perhaps NORMAL_STREAMING_MAX_DELAY?  Also, doesn't this const already exist somewhere? How is this duplication going to be resolved?

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+import random
+import time
+from enum import Enum
+from typing import Optional, Protocol
+
+from ldclient.impl.util import log
+
+# The delay bounds of the extended regime, in seconds. A component enters the
+# extended regime after an unexpected failure.
+EXTENDED_INITIAL_DELAY = 5 * 60
```
tanderson-ld, line 37: Is it the norm in python that all time values / APIs are float seconds?

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+# How many polls in a row must succeed before polling's retry state resets.
+POLLING_RESET_SUCCESSES = 2
+
+# HTTP statuses in the 4xx range that are still normal failures. Every other
+# 4xx is unexpected.
+_NORMAL_4XX_STATUSES = frozenset([400, 408, 429])
+
+# An upper bound on the backoff exponent, so a long outage cannot overflow the
+# delay computation. Any real ceiling is reached long before this.
+_MAX_BACKOFF_EXPONENT = 30
```
tanderson-ld, line 53: This approach for calculation can remove the need for bounds protection logic.

https://github.com/launchdarkly/dotnet-eventsource/pull/126/changes#diff-16849bf43eb8418d2b4129594604eba65613dabf4f40abc7f6dae553239de81cR65

https://github.com/launchdarkly/dotnet-eventsource/pull/126/changes#diff-16849bf43eb8418d2b4129594604eba65613dabf4f40abc7f6dae553239de81cR86

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+        :param normal_ceiling: the longest normal-regime delay, in seconds
+        :param extended_initial_delay: the delay before the first retry in the
+            extended regime, in seconds
+        :param extended_ceiling: the longest extended-regime delay, in seconds
+        :param reset_policy: decides when the retry state resets
+        :param operating_cadence: the rate the component normally operates at,
+            in seconds; no wait is ever shorter than this. Zero disables the
+            floor, which is what streaming wants.
+        """
+        self._initial_delay = initial_delay
```
tanderson-ld: Since this is greenfield, might as well call this normal_initial_delay.

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+        self._normal_ceiling = normal_ceiling
+        self._extended_initial_delay = extended_initial_delay
+        self._extended_ceiling = extended_ceiling
+        self._reset_policy = reset_policy
+        self._operating_cadence = operating_cadence
```
tanderson-ld, line 180: It seems like this operating_cadence is to help with the normal second poll after the first success? Can't this be calculated based on normal_initial_delay?  It seems like redundant state?  Maybe I'm not understanding the usage of operating_cadence and initial_delay set to different non-zero values.

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+        Records a successful operation, and resets the retry state if that is
+        now enough.
+
+        The wait before the next operation becomes the operating cadence, even
+        when the retry state is still raised, because a backoff wait applies to
+        a retry and not to every operation.
+        """
+        self._reset_policy.note_healthy()
+        self._reset_if_due()
+        self._next_delay = self._operating_cadence
```
tanderson-ld, line 237: This can be 0? I think this indicates a logical flaw in how the values are flowing through / overriding each other.

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+        delay = min(self._min_delay * (2**exponent), self._max_delay)
+        jitter = random.random() * delay / 2
+        return max(delay - jitter, self._operating_cadence)
+
+
+def for_streaming(initial_reconnect_delay: float) -> RetryState:
+    """
+    Builds the retry state for a streaming data source.
+
+    Streaming has no operating cadence, so there is no floor on the wait. It
```
tanderson-ld: `so there is no floor on the wait` .  We do specify a floor of 1 second on streaming attempts.  Is this referring to something else?

ldclient/impl/retry.py
```diff
@@ -0,0 +1,337 @@
...
+    @property
+    def operating_cadence(self) -> float:
+        """The rate the component normally operates at, in seconds."""
+        return self._operating_cadence
+
+    @property
+    def in_extended_regime(self) -> bool:
+        """Whether an unexpected failure has moved this state to the extended
+        delay bounds."""
+        return self._extended
```
tanderson-ld: Many of these properties aren't exposed in other implementations and I fear only bad things can come of some of these being exposed.  Other code will try to attach logic to these properties and they seem like they could have race conditions and be hard to prove are well behaved when used.

ldclient/impl/retry.py
```diff
@@ -288,6 +294,19 @@ def _compute_wait(self) -> float:
...
+def _positive_finite(value: float, default: float, name: str) -> float:
+    """Returns ``value`` if it is a positive, finite number of seconds, and the
+    default otherwise. A non-finite value would make the jitter arithmetic
+    produce a NaN delay, and a non-positive one would retry with no wait."""
+    if value > 0 and math.isfinite(value):
+        return value
+    log.warning(
+        "%s must be a positive, finite number of seconds; using the default of %ss"
+        % (name, default)
+    )
```
tanderson-ld, line 259: Are logger instances ever passed around in the code base?  This static call stands out to me coming from our other code bases.

jsonbailey: No, its is the standard in this repo to pull from the import. And it is different then most other languages.

ldclient/impl/retry.py
jsonbailey: You can think of operating cadence as the time between normal healthy operations and the normal_initial_delay as the starting point for delays between failed operations. For polling, these are the same, but for streaming the default cadence is 0s, but the failed starts at 1s or whatever the user configures so that a backoff can actually be calculated.

ldclient/impl/retry.py
jsonbailey: Yes. `time.monotonic()`, `time.sleep()`, `threading.Event.wait()` and `asyncio.sleep()` all take or return float seconds, so a float is what every call site here needs. `Config` already exposes `poll_interval` and `initial_reconnect_delay` as float seconds too, so this matches what callers already pass in.

ldclient/impl/retry.py
jsonbailey: I have removed the duplication between this and the config (it was using hardcoded values vs a constant). There are additional duplicates in FDv2, but those will be removed when we update them later to use the new RetryState.

ldclient/impl/retry.py
jsonbailey: We are limited by the float seconds data type that is used in the SDK so those bound protections won't work.

ldclient/impl/retry.py
jsonbailey: It can be 0, we can meet if it would be helpful to talk through this but 0 is intentional (for streaming).

tanderson-ld (approved):

